PR #21, opened 2022-10-06
Fix/drawer

goncalosard:
## Summary

Fixing the bug on the Drawer component inside the package core where if you closed the Drawer using the button, it would break, and never open again correctly. The fix contains a condition so it allows the closing using the button to happen correctly.

Tested locally.


danielleroux (approved):
lgtm

PR #33, opened 2022-10-10
fix(core/toast): set default icon to info

nuke-ellington:
Set default icon for toasts to info


danielleroux (changes requested):
Headline with brand-theme looks not aligend

danielleroux (commented):
Headline looks not aligend in brand-theme

danielleroux (changes requested):
Headline looks not aligend in brand-theme

danielleroux (approved):

Issue #16, opened 2022-10-05
documentation issue list


nuke-ellington:
[] breadcrumb separator SVG color does not change in dark theme
[] remove edit this page button
[] edit next buttons
[] theme page rework (additional theme)
[] color table (custom property names are editable)
[] find better and globally consistent dummy data (e.g. Porsche, Toyota)

ChristianButter:
<img width="1115" alt="Bildschirmfoto 2022-10-06 um 15 23 27" src="https://user-images.githubusercontent.com/115155128/194324457-9c6bc357-662e-4254-ba4c-acaa900cf4cb.png

Datepicker must have a dark shadow

ChristianButter:
<img width="654" alt="Bildschirmfoto 2022-10-06 um 15 25 07" src="https://user-images.githubusercontent.com/115155128/194324911-e3d67efd-e38c-41d4-9a8c-492f319d3627.png">
<img width="654" alt="Bildschirmfoto 2022-10-06 um 15 25 07" src="https://user-images.githubusercontent.com/115155128/194324928-a2697aef-672b-4b1b-985c-e0902de4003a.png">

Lower right corner of DateTimePicker layout not correct.

ChristianButter:
<img width="1105" alt="Bildschirmfoto 2022-10-06 um 15 29 01" src="https://user-images.githubusercontent.com/115155128/194325866-898993c9-0ab8-459e-8d0f-8568eb520411.png">
Legend text not readable in dark mode

Also check font color of axis in dark mode. seems to dark

ChristianButter:
<img width="1212" alt="Bildschirmfoto 2022-10-06 um 15 36 12" src="https://user-images.githubusercontent.com/115155128/194327366-58267710-39ea-4de1-9218-0cb1bfc3fd20.png">
Modal not openning when button is pressed

ChristianButter:
<img width="588" alt="Bildschirmfoto 2022-10-06 um 15 41 02" src="https://user-images.githubusercontent.com/115155128/194328388-20c0cdf1-555f-4425-89f7-53bdf7bd66d0.png">
Toast does not open when button is pressed

silviowolf:
![image](https://user-images.githubusercontent.com/115155312/194339012-429d76c4-e083-4051-a72e-68955a6e7922.png)
Styling of eventlist should be checked (border, background, States: hover, active, selected, selected-hover, selected-active)

silviowolf:
[...]

PR #52, opened 2022-10-12
Fix/about and legal

goncalosard:
## Summary
Fixed problems from documentation issue list #16

The fixed topic were:
- about and legal is visible outside of iframe when opening/closing
- about and legal close button is partially outside of content area
- about and legal - no initial tab selection

Also fixed the following topic not from the about and legal but from the Color file:
- color table (custom property names are editable)

Tested locally


packages/core/src/components/menu-about/menu-about.tsx
```diff
@@ -34,8 +34,7 @@ export class MenuAbout {
   /**
    * Active tab
    */
-  @Prop({ mutable: true, reflect: true }) activeTabLabel: string =
-    this.i18nImprintLabel;
+   @Prop({ mutable: true }) activeTabLabel: string;
```
danielleroux, line 37: If we want to remove this prop we should mark the i18nImprintLabel as deprecated or remove it complete because i think this is an leftover from the refactoring

danielleroux (changes requested):

danielleroux (approved):

PR #70, opened 2022-10-18
feat(documentation): mobile version

goncalosard:
## Summary

Implemented the mobile version for the 428 resolution

Tested locally


danielleroux (changes requested):
Tested only with the chrome dev tools but the inital screen looks still noch aligend correct
image

Maybe we can disable the Intro animation for mobile devices? @ChristianButter

ChristianButter:
@danielleroux @goncalosard
To fix mobile issues a quickly as possible, I would also disable the animation on mobile devices. But not on other screen resolutions.

danielleroux (approved):

PR #71, opened 2022-10-18
Fix/icon library

goncalosard:
## Summary

Changed the search on the icons page to not be case-sensitive.

Tested locally


danielleroux (approved):
lgtm

PR #96, opened 2022-10-27
fix(core/expanding-serach): expanding search sizing

goncalosard:
## Summary

Fixed the problem where in collapsed state the component had the size of expanded state.

Now the size when collapsed is the size of the icon only.

With this new change, it meant there was a problem where if it all the way to the right(ex: flex-end), the component would overflow outside of the div it was in. Also fixed this problem.


danielleroux (approved):
- Changed the flex to inline-flex.
- Added a visual regression-testcase

lgtm.

PR #99, opened 2022-10-28
Fix(core): menu close overlays

goncalosard:
## Summary

It was changed to when a menu item is selected, the menu closes, it also closes the about and legal & setting content.


packages/core/src/components/menu-about/menu-about.tsx
```diff
@@ -46,7 +47,7 @@ export class MenuAbout {
   /**
    * About and Legal closed
    */
-  @Event() close: EventEmitter<MouseEvent>;
+  @Event() close: EventEmitter;
```
danielleroux, line 50: This cause a major change in our component api. Is there any reason behind it?

packages/core/src/components/menu-item/menu-item.tsx
```diff
@@ -49,6 +49,11 @@ export class MenuItem {
 
   @State() title: string;
 
+  /**
+   * Event to emit to parent that the item was selected
+   */
+  @Event() itemClicked: EventEmitter<boolean>;
+
```
danielleroux, line 56: Is a new event necessary? Why not the default click event?

danielleroux (changes requested):

danielleroux (approved):

nuke-ellington (approved):

PR #116, opened 2022-11-04
fix: format code and markup

sisp:
## Summary

While scanning the code a bit, I noticed it's not consistently formatted using Prettier, although Prettier is used as the code formatter. So I ran Prettier on the entire repo with some ignored files and directories mainly taken from `.gitignore`.

In some `.js`/`.ts[x]` files, the imports weren't sorted properly, so I added the the package [`prettier-plugin-organize-imports`](https://www.npmjs.com/package/prettier-plugin-organize-imports) which makes Prettier format imports using the `organizeImports` feature of the TypeScript language service API. I've used it in other projects in the past and it's a great plugin.

While running Prettier, I saw it choking on

https://github.com/siemens/ix/blob/67ff00837402b6041c117f40146bcaff62a37b85/packages/core/src/components/select/test/mode-multiple/index.html#L16

which I've fixed (`<bodyp ...>` -> `<body>`) along the way, too.

Sorry for the huge PR, but it's just tons of Prettier formatting changes.

There's not much to test, just formatting changes. :crossed_fingers: that the CI checks pass. :wink:


sisp:
It seems the build generates a bunch of Markdown files and also formats some JS/TS code. From what I can tell, Stencil, [`packages/documentation/scripts/generate-api.mjs`](https://github.com/siemens/ix/blob/67ff00837402b6041c117f40146bcaff62a37b85/packages/documentation/scripts/generate-api.mjs), and possibly some other scripts are generating those files. Shouldn't those generated files be either ignored by Git? :thinking:

danielleroux:
> It seems the build generates a bunch of Markdown files and also some JS/TS code. From what I can tell, Stencil, [`packages/documentation/scripts/generate-api.mjs`](https://github.com/siemens/ix/blob/67ff00837402b6041c117f40146bcaff62a37b85/packages/documentation/scripts/generate-api.mjs), and possibly some other scripts are generating those files. Shouldn't those generated files be ignored by Git? :thinking:
> 
> 

good point. I've also already thought about that.

I think put autogenerated code under version control is not a bad idea at all. (js/ts)

- You see changes directly in the diff. Some wird autogenerated bugs (not often but happens)
- If some one checkout the repository and install the node_modules he is ready so browse the code without missing parts (this is also a point for GitHub browser view)

But for the markdown files maybe they can be part of the gitignore file. At least for documentation/docs/Auto-generated because this files are a markdown output of the component-doc.json (packages/core)

sisp:
Alright, I can probably agree on versioning the auto-generated JS/TS code. But then it should be formatted using Prettier before being committed. I doesn't seem that Stencil supports some kind of post-generation command/hook for running Prettier. Perhaps an alternative solution is to use a pre-commit hook, e.g. a setup using Husky + [lint-staged](https://github.com/okonet/lint-staged) might help.

Regarding the auto-generated docs, I've created a PR that removes them and adds the path to `.gitignore`: #117

sisp:
I guess one problem with versioning generated code is that you might end up with dead files at some point when the code generation changes and you don't remember to manually delete the obsolete files.

danielleroux:
Will be intergrated with #197

PR #118, opened 2022-11-05
chore: move `.gitignore` rules to packages

sisp:
## Summary

I've moved some `.gitignore` rules into package-local `.gitignore` files which IMO makes keeping track of the relevant rules easier than managing them globally.

Just one question: It seems `packages/documentation/static/versioned_examples/` is populated ever. Is that right? The `.gitignore` rule

```gitignore
!packages/documentation/static/versioned_examples/*/webcomponent-examples
```
since none of its parent directories was excluded, so there's no need to _include_ this specific subpath again.

Nothing to test here.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -130,7 +130,6 @@ Thumbs.db
 /build
 
 # Generated files
-.docusaurus
 .cache-loader
 
 # Misc
@@ -152,16 +151,3 @@ packages/*/www/
 packages/*/loader/
 packages/*/dist-transpiled/
 packages/*/build/
-
-###
-# @siemens/html-test-app
-###
-
-packages/html-test-app/src/public/additional-theme
-
-###
-# documentation
-###
-
-packages/documentation/static/webcomponent-examples/*
-!packages/documentation/static/versioned_examples/*/webcomponent-examples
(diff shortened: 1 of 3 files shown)


danielleroux:
@sisp It was intended to keep only one gitignore file. Can you please move the lines back to the root/.gitignore.

```
Just one question: It seems packages/documentation/static/versioned_examples/ is populated ever. Is that right? The .gitignore rule

!packages/documentation/static/versioned_examples/*/webcomponent-examples
```
Yes we use the docusaurus versioning. (https://docusaurus.io/docs/versioning - https://github.com/siemens/ix/blob/main/packages/documentation/src/components/Demo.tsx#L42)

danielleroux (changes requested):
@sisp It was intended to keep only one gitignore file. Can you please move the lines back to the root/.gitignore.

```
Just one question: It seems packages/documentation/static/versioned_examples/ is populated ever. Is that right? The .gitignore rule

!packages/documentation/static/versioned_examples/*/webcomponent-examples
```
Yes we use the docusaurus versioning. (https://docusaurus.io/docs/versioning - https://github.com/siemens/ix/blob/main/packages/documentation/src/components/Demo.tsx#L42)

sisp:
I'll just close the PR if you prefer a single `.gitignore` file at the root of the project; nothing to move back.

PR #129, opened 2022-11-09
fix(core/dropdown): breadcrumb dropdown overflow

### LINKED ISSUE #77: Breadcrumb dropdown does not have max height
## I'm submitting a...

<pre><code>
[x] Bug report
[ ] Feature request
</code></pre>

## Current behavior

The dropdown of the breadcrumb will grow past the screen borders if there are too many items inside.

## Expected behavior

Both max height and overflow detection relative to the browser viewport should prevent items that are not accesible.

goncalosard:
## Summary

Breadcrumb dropdown now has the max-height for 5 items, after that it will have a scroll for the overflow.

Fixes #77  

Tested locally


packages/core/src/components/breadcrumb/breadcrumb.scss
```diff
@@ -186,7 +186,7 @@
   @include breadcrumb('ghost');
 }
 
-.dropdown-overflow .dropdown-menu {
-  max-height: 25rem;
```
nuke-ellington, line 190: The dropdown will still grow past the screen height if it contains too many items.

danielleroux: max height should something like `70vh`

danielleroux (changes requested):
Please add a visual regression testcase

goncalosard:
> Please add a visual regression testcase

@danielleroux The fix is done and the visual regression test is done. Its not passing the build only here, on my machine works perfect. I think its because it is expecting to have a snapshot from a linux, you should just run the tests, and commit the snapshot.

danielleroux:
> > Please add a visual regression testcase
> 
> @danielleroux The fix is done and the visual regression test is done. Its not passing the build only here, on my machine works perfect. I think its because it is expecting to have a snapshot from a linux, you should just run the tests, and commit the snapshot.

Did you executed the test inside Docker?

I will update the snapshots.

danielleroux:
@goncalosard I checked the PR i will refactor the changes regarding the breadcrumb a little bit, because it is more sustainable to adapt the max-height within the dropdown it self

nuke-ellington (approved):

danielleroux:
@nuke-ellington @goncalosard Can you please review the changes, i have made regarding dropdown?

I will also resolve the merge conflicts

danielleroux (approved):

PR #147, opened 2022-11-17
fix(core/menu-item): deprecate bottom property in favor of slot based implementation

nuke-ellington:
## Summary

Since the old implementation using the bottom property on menu items had some problems and will not work anymore please use slot="bottom" instead.

- [x] Added visual regression test


danielleroux (approved):
lgtm

Changed the title regarding squash merge

PR #148, opened 2022-11-17
fix(core): eol uniformization

PhentomPT:
## Summary

When running **build** for the _@siemens/ix_ package the file **component-doc.json** generates its content _eol_ based on the users system.
To uniformize the content from win/unix systems i added a search replace "\r" -> ""

Runned the build command on both unix & win systems and the output is the same


danielleroux (approved):
lgtm

PR #153, opened 2022-11-21
refactor(core): replace popperjs with floating-ui

danielleroux:


packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -81,12 +88,22 @@ export class Dropdown {
    */
   @Prop() header?: string;
 
+  /**
+   * Move dropdown along main axis of alignment
```
nuke-ellington, line 92: Rather unclear description :)

packages/core/src/components/dropdown/placement.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export declare type BasePlacement =
+  | typeof top
+  | typeof bottom
+  | typeof right
+  | typeof left;
+
+export declare type AutoPlacement = 'auto' | 'auto-start' | 'auto-end';
+
+export declare type PlaceMentWithAlignment =
```
nuke-ellington, line 22: Typo (capital M)

packages/core/src/components/menu-avatar/menu-avatar.tsx
```diff
@@ -106,7 +85,14 @@ export class MenuAvatar {
             </span>
           </div>
         </li>
-        <ix-dropdown show={this.displayMenu}>
```
nuke-ellington, line 109: Looks like semver: major to me

nuke-ellington (changes requested):

packages/core/src/components/dropdown/dropdown.tsx
danielleroux: https://floating-ui.com/docs/offset#options

Any other suggestions?

packages/core/src/components/menu-avatar/menu-avatar.tsx
danielleroux: Do you marked a wrong line?

packages/core/src/components/dropdown/dropdown.tsx
nuke-ellington: I understand this is the floating UI API but it seems a little strange to have this object hash with only one property.

danielleroux: Its only an internal prop its not exposed in documentation in some way. I can copy the complete obejct definition?

I dont want to reuse the typing of the floating lib because then we have a technical dependency in our components definition which is also not the best

nuke-ellington: If this is the only property we are using and it is not part of the API I think we can leave it like this.

nuke-ellington (approved):

nuke-ellington (approved):

nuke-ellington (approved):

PR #163, opened 2022-11-22
fix(documentation): toast missing properties

### LINKED ISSUE #157: Toast: missing properties in documentation
## I'm submitting a...

<pre><code>
[x] Bug report
[ ] Feature request
</code></pre>

## Current behavior

Documentation of the toast properties seems to be missing.  
 https://ix.siemens.io/docs/controls/toast

goncalosard:
## Summary

Added missing properties to Toast page on documentation.

Fix #157 

Tested locally

### DIFF
--- a/packages/documentation/docs/controls/toast.md
+++ b/packages/documentation/docs/controls/toast.md
@@ -1,5 +1,6 @@
 import Preview from '@site/src/components/Preview';
 import TabItem from '@theme/TabItem';
+import Props from './../auto-generated/ix-toast/props.md';
 
 import SourceToast from './../auto-generated/previews/web-component/toast.md';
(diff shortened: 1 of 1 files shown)


danielleroux:
@goncalosard We don't need add the toast props here. The issue is in Jour Fixe it is not intended to take issues labeled with "JourFix" because these issue need additional attension.

Regard the issue #157 in this case we need a better implementation for the toast-utils (https://github.com/siemens/ix/blob/main/packages/core/src/components/toast/toast-utils.ts)

This implementation is different for each framework 
- angular: https://github.com/siemens/ix/blob/main/packages/angular/src/toast/toast.service.ts
- react: https://github.com/siemens/ix/blob/main/packages/react/src/toast/toast.ts

PR #169, opened 2022-11-23
fix(core/map-navigation): use ix-application-header for header part

danielleroux:
## Summary

Reuse `ix-application-header` inside map-navigation

Update visual regression snapshots


nuke-ellington (changes requested):
App name seems too close to icon slot. Siemens logo f.e. will not look good like that.

danielleroux:
> App name seems too close to icon slot. Siemens logo f.e. will not look good like that.

@nuke-ellington 
image

nuke-ellington (approved):
Does not meet design spec fully but should not be problematic.

PR #185, opened 2022-11-25
fix(core/event-list): re-enable chevron and prevent event list from h…

nuke-ellington:
…iding slotted icons

VRT


danielleroux (approved):
lgtm

PR #187, opened 2022-11-25
fix(core): checkbox changes to match design

### LINKED ISSUE #176: checkbox size is not aliged with design spec
## I'm submitting a...

<pre><code>
[x] Bug report
[ ] Feature request
</code></pre>

## Current behavior

Checkbox size is not aligned with design spec

goncalosard:
## Summary

Fix #176 

Changed the sizes to match zeplin design.
Please check if this was the intended result.

Tested locally.


packages/core/scss/components/_checkboxes.scss
```diff
@@ -40,8 +42,8 @@
 
   [type='checkbox'] + label:before {
     content: '';
-    width: $size;
-    height: $size;
+    width: 18px;
```
danielleroux, line 45: @goncalosard same as with radiobutton we check that the `$size` variable should be updated instead of the inline px value

danielleroux (changes requested):

danielleroux (changes requested):
@goncalosard `Indeterminate` state looks wrong

image

We should also a visual-regression test for it.

danielleroux (approved):
lgtm

nuke-ellington (approved):

PR #461, opened 2023-03-22
fix(core/blind): improve keyboard accessibility

jg0815:
## Pull request checklist

- [x] Unit tests (`yarn test`) were run locally and passed
- [ ] Visual Regression Tests (`yarn visual-regression`) were run locally and passed
- [ ] Linting (`npm lint`) was run locally and passed

## Pull request type

- [ ] Bug fix
- [x] Feature
- [ ] Build related changes

## What is the new behavior?

Prior to this change, blind components were not keyboard operable, which was an accessibility issue. This has now been fixed, while also providing the required visibility of the associated focus frame. To make the individual elements of blind component instances referenceable within the required WAI-ARIA attribution, they have been assigned unique IDs, which has been realized via an additional module-scoped sequential number (namely 'sequentialInstanceId').

- [ ] Yes
- [x] No


packages/core/src/components/blind/blind.scss
```diff
@@ -21,6 +21,10 @@
   border-radius: var(--theme-blind--border-radius);
   overflow: hidden;
 
+  :focus {
```
nuke-ellington, line 24: In general we try to use focus-visible instead of focus. Would this lead to issues with regards to a11y?
Also we have some SASS mixins defined for that:
@import 'mixins/hover';
[...]
@include focus-visible {
[...]
}

jg0815: This setting was made because the already existing SASS mixin (@include focus-visible) obviously shows the focus frame incompletely and in the wrong color. However, if this is actually desired in the previous way, the ":focus" directive can of course be removed again.

nuke-ellington: If `focus-visible` has the wrong color we need to change that. Would it be acceptable from an ally perspective to switch to `focus-visible`?

jg0815: As long as it is guaranteed that the focus frame is rendered adequately, completely and in the valid color, there is really nothing wrong with doing this from an a11y perspective.

packages/core/src/components/blind/blind.tsx
danielleroux, line 1: We should use format strings e.g.

```
aria-labelledby={`ix-blind-header-title-${this.id}`}
```

danielleroux (changes requested):

danielleroux (approved):

packages/core/src/components/blind/blind.tsx
```diff
@@ -129,12 +125,9 @@ export class Blind {
             'blind-header': true,
             closed: this.collapsed,
           }}
-          type="button"
-          role="button"
-          tabindex="0"
           aria-labelledby={`ix-blind-header-title-${this.id}`}
           aria-controls={`ix-blind-content-section-${this.id}`}
-          aria-expanded={this.collapsed}
+          aria-expanded={`${!this.collapsed}`}
```
danielleroux, line 130: use binary value please i changed it wrong in the last commit `{!this.collapsed}`

packages/core/src/components/blind/blind.tsx
```diff
@@ -129,12 +125,9 @@ export class Blind {
             'blind-header': true,
             closed: this.collapsed,
           }}
-          type="button"
```
danielleroux: please do not remove type='button' otherwise some browser will use the button as form submit

danielleroux (changes requested):

packages/core/src/components/blind/blind.tsx
danielleroux: fixed

packages/core/src/components/blind/blind.tsx
danielleroux: added

Issue #509, opened 2023-05-03
Corner radius for modals


flxlst09:
There is a discrepancy between the corner radius for Modals:

Figma: 4px
Implemented: 0.5rem

<img width="940" alt="modaldialog_cornerradius" src="https://user-images.githubusercontent.com/116285506/235901281-40090982-2646-478e-a29b-e5247bc7c64c.png">

silviowolf:
Thanks, you're right. should be 0.25 rem instead.
Will be fixed.

silviowolf:
0.5rem are coming from Bootstrap vaiables

PR #517, opened 2023-05-05
fix(core/modal): border-radius value

goncalosard:
## Pull request checklist

- [ ] Unit tests (`yarn test`) were run locally and passed
- [x] Visual Regression Tests (`yarn visual-regression`) were run locally and passed
- [ ] Linting (`npm lint`) was run locally and passed

## Pull request type

- [x] Bug fix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: #509 

## What is the new behavior?

- Changed border-radius value to correct value from Figma

- [ ] Yes
- [x] No


packages/core/src/components/modal/modal.scss
```diff
@@ -56,6 +56,7 @@
         var(--theme-modal--border-color);
       padding: $large-space;
       max-height: 100vh;
+      border-radius: 0.25rem;
```
danielleroux, line 59: We should use `var(--theme-default-border-radius)`

danielleroux (changes requested):
We should also try to improve the visual regression test first, because the test has to fail without updating the image refs.

Maybe it is necessary to adapt the screenshot target area to reduce the compared pixel ratio.

danielleroux (approved):

PR #845, opened 2023-10-12
docs: website improvements

mfruehmannext:
## Other information

fixes a couple of small issues regarding the website; adds scss support for website's stylesheets.


packages/documentation/src/components/Card.module.css
```diff
@@ -23,21 +23,51 @@
...
+    width: 100%;
+    margin-right: 0;
+  }
+
+  @media (min-width: 664px) and (max-width: 684px) {
+    max-width: 640px;
+  }
+
+  @media (min-width: 997px) and (max-width: 1087px) {
+    /*margin-right: 0;*/
```
danielleroux, line 39: Is the code still necessary? If not please remove.

packages/documentation/docusaurus.config.js
```diff
@@ -35,9 +35,9 @@ try {
 const customCss = [
   ...libCss,
   require.resolve('./src/css/custom.css'),
-  require.resolve('./src/css/search.css'),
-  require.resolve('./src/css/api-table.css'),
-  require.resolve('./src/css/cards.css'),
+  require.resolve('./src/css/search.scss'),
+  require.resolve('./src/css/api-table.scss'),
+  require.resolve('./src/css/cards.scss'),
```
danielleroux, line 40: Why renamed to scss if there are not changes inside the styles files related to scss?

danielleroux (changes requested):

packages/documentation/docusaurus.config.js
mfruehmannext: because imho it does not make any sense to use css and scss in combination. but - your choice.

packages/documentation/src/components/Card.module.css
mfruehmannext: removed in latest commit

packages/documentation/docusaurus.config.js
danielleroux: But you didn't changed `custom.css` then

danielleroux (approved):

PR #870, opened 2023-10-23
fix(core/modal): ensure that key down listener gets called

### LINKED ISSUE #848: Modal: not possible to disable close function by clicking on ESC
### What happened?

I tried to set the `keyboard ` config to false, but I can still use ESC button to close the modal if I click on that.

User experience: 
1. Open the modal window
2. Press ESC -> nothing happens (as expected)
3. Click on the dialog(!!!) then press ESC -> modal is closed

Note: I couldn't find description about the `keyboard` config in the documentation, so it is just an assumption that this is how it should work. 

Angular

v2.0.3

```shell
this.modalService.open({
 content: DialogComponent, keyboard: false
});
```

nuke-ellington:
## Pull request checklist

- [x] Unit tests (`yarn test`) were run locally and passed
- [ ] Visual Regression Tests (`yarn visual-regression`) were run locally and passed
- [x] Linting (`npm lint`) was run locally and passed

## Pull request type

- [x] Bug fix
- [ ] Feature
- [ ] Build related changes

GitHub Issue Number: Closes #848

## What is the new behavior?

-
-
-

- [ ] Yes
- [x] No


packages/core/src/components/modal/modal.tsx
```diff
@@ -86,6 +87,13 @@ export class Modal {
    */
   @Event() dialogDismiss: EventEmitter;
 
+  @OnListener<Modal>('keydown', (self) => !self.keyboard)
```
danielleroux, line 90: Why is global event listener here necessary?

danielleroux (changes requested):
Component test is missing

packages/core/src/components/modal/modal.tsx
nuke-ellington: Because the onKey listener will only get called when the dialog has focus. Clicking on it removes the focus.

danielleroux (approved):

PR #1196, opened 2024-04-05
fix(core/select): arrow key navigation

nuke-ellington:

GitHub Issue Number: IX-933

-
-

## 🏁 Checklist


packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -194,10 +203,12 @@ export class Dropdown implements ComponentInterface, DropdownInterface {
     return this.hostElement.shadowRoot.querySelector('slot');
   }
 
-  private disposeListener?: Function;
+  private disposeClickListener?: Function;
+  private disposeKeyListener?: Function;
```
danielleroux, line 207: Avoid the `Function` type, as it provides little safety for the following reasons:
It provides no type safety when calling the value, which means it's easy to provide the wrong arguments.
It accepts class declarations, which will fail when called, as they are called without the new keyword.
([Source](https://typescript-eslint.io/rules/ban-types/))

e.g.

```
disposeClickListener?: () => void;
```

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -316,6 +349,24 @@ export class Dropdown implements ComponentInterface, DropdownInterface {
...
+
+      this.itemObserver.observe(this.dropdownRef, {
+        childList: true,
+        subtree: true,
+      });
+
+      this.registerKeyListener();
+    } else {
+      this.arrowFocusController.disconnect();
+      this.arrowFocusController = undefined;
```
danielleroux, line 367: Not needed here

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -405,6 +456,12 @@ export class Dropdown implements ComponentInterface, DropdownInterface {
     );
   }
 
+  private focusDropdownItem(index: number) {
+    requestAnimationFrame(() => {
+      this.dropdownItems[index]?.shadowRoot.querySelector('button').focus();
```
danielleroux, line 461: DropdownItem's implements DropdownItemWrapper interface which gives you access to the `getDropdownItemElement` maybe a better solution as query the shadowRoot?

packages/core/src/components/select/select.scss
```diff
@@ -69,16 +69,30 @@
...
       outline: none;
     }
   }
 
+  .select.disabled {
+    border-bottom: var(--theme-input--border-thickness) solid
+      var(--theme-color-weak-bdr);
+
+    input {
+      color: var(--theme-color-weak-text, rgba(255, 255, 255, 0.45));
```
danielleroux, line 83: Do we need the fallback `rgba(255, 255, 255, 0.45)`?

packages/core/src/components/select/select.tsx
```diff
@@ -127,24 +129,52 @@ export class Select {
   @State() dropdownWrapperRef!: HTMLElement;
   @State() dropdownAnchor!: HTMLElement;
   @State() isDropdownEmpty = false;
-  @State() hasFocus = false;
-  @State() navigationItem: HTMLIxSelectItemElement;
+  @State() navigationItem: HTMLIxSelectItemElement | HTMLIxDropdownItemElement;
```
danielleroux, line 132: We could use here also the `DropdownItemWrapper`
```
class SelectItem implement DropdownItemWrapper {
...
}
```

packages/core/src/components/select/select.tsx
```diff
@@ -338,12 +409,17 @@ export class Select {
...
+    let item;
+
     if (this.editable && !this.itemExists(this.inputFilterText)) {
       this.emitAddItem(this.inputFilterText);
-      this.navigationItem = this.items[this.items.length - 1];
+      item = this.items[this.items.length - 1];
+    }
+
+    if (item) {
+      item['onItemClick']?.();
```
danielleroux, line 420: Define a function (annotate it with `@Method`) on the component and call it async.

danielleroux (changes requested):

packages/core/src/components/dropdown/dropdown.tsx
nuke-ellington: Need to focus the nested button for the focus-visible state to work

danielleroux (approved):

PR #1214, opened 2024-04-11
fix(core/select-item): check by undefined not by value

jul-lam:

0 is currently not a valid value option for the ix-select-item.

GitHub Issue Number: xxx

0 is now a valid value option for the ix-select-item.

## 🏁 Checklist

## 👨‍💻 Help & support

The test is skipped, since html only uses string values, which the value prop also will use in 3.0.

### DIFF
--- a/.changeset/sweet-yaks-appear.md
+++ b/.changeset/sweet-yaks-appear.md
@@ -1,0 +1,5 @@
+---
+"@siemens/ix": patch
+---
+
+fix(select-item): make 0 a valid value option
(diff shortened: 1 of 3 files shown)


nuke-ellington (approved):

nuke-ellington (commented):
Do we have a test case?

jul-lam:
@nuke-ellington - I haven’t set up a test because HTML exclusively uses string values for properties. We could however set up a test using React, but I was not sure if that’s needed, since our value property will take exclusively string in our next major release.

PR #1219, opened 2024-04-16
fix(core/workflow): add flexible layout

### LINKED ISSUE #1090: Workflow steps don't scale with their content
### Prerequisites

- [X] I have not leaked any internal company information.
- [X] I have not posted internal links or code snippets.

### What happened?

I used the workflow component inside a smaller div which was scaled to only use 40% of the whole screen. All my workflow steps then overflowed out of the screen. I found that the step has a predefined width which made it not scale to fit in my small div:  https://github.com/siemens/ix/blob/0c13d91562017bd3b5773fb3976b41d69fa037ab/packages/core/src/components/workflow-step/workflow-step.scss#L25

Also the property width is defined twice:
https://github.com/siemens/ix/blob/0c13d91562017bd3b5773fb3976b41d69fa037ab/packages/core/src/components/workflow-step/workflow-step.scss#L22

My workaround was to override this with in my code:
```css
// Cannot use step, because component uses .step already as class and then the changes don't get applied
.stepElement > div {
    width: auto !important;
    height: auto !important;
}
```

I would have expected the steps to scale so that they fit inside their parent container without getting too small.

React

v1.6.3

```shell
https://stackblitz.com/edit/react-lpbaev?file=src%2FApp.js
```

### LINKED ISSUE #1194: Ix-workFlow steps having fixed width
### Prerequisites

### What happened?

Hi Team,

In Ix-workFlow steps, In below UI, we need to provide the width till the end of that white card. But we can't able to increase/alter the width of ix-workflow steps. So what will be the recommended solution for this? 

 

Do we need to decrease the white card width or is there a way to increase the width of the workflow step?

image

Angular

2.0.4

```shell
<ix-workflow-steps vertical>
    <ix-workflow-step status="done">Sample text 1</ix-workflow-step>
    <ix-workflow-step status="success">Sample text 2</ix-workflow-step>
    <ix-workflow-step status="open">Sample text 3</ix-workflow-step>
    <ix-workflow-step status="warning">Sample text 4</ix-workflow-step>
  </ix-workflow-steps>
```

matthiashader:
- ix-workflow-steps and also the single step component having a fixed with and height
- that prevents the user from putting this component in a container with a small width and height

GitHub Issue Number: Closes #1090, Closes #1194, [IX-936] 

- the component has been adapted to be more flexible
- horizontal -> takes the available width
- vertical -> takes the available height
- according to @silviowolf the height of the steps should not be restricted, this should be done by the parent container.

## 🏁 Checklist


.changeset/wise-games-relate.md
```diff
@@ -0,0 +1,5 @@
+---
+'@siemens/ix': minor
```
danielleroux, line 2:
```suggestion
'@siemens/ix': patch
```

should be a patch not a minor change

danielleroux (changes requested):

danielleroux (approved):

Issue #1207, opened 2024-04-10
 IxGroupItem should have pointer cursor


tomas-reznak:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [X] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

I think cursor should be adjusted for group items to match it's purpose (as a link to sub item, esp. when using button element inside). The only exception is when group item is set as inactive (by supressSelection prop).

Current wrong behavior:

![group-item-cursor-wrong](https://github.com/siemens/ix/assets/99844553/08443896-5ad7-4602-9070-b6a545e42a03)

Correct behavior:

![group-item-cursor-correct](https://github.com/siemens/ix/assets/99844553/e115fa37-3ec7-41c7-8e62-57f5ac4a3585)

### What type of frontend framework are you seeing the problem on?

React

### Which version of iX do you use?

v2.1.2

### Code to produce this issue.

```shell
<IxGroup header="Group header" subHeader="Group subheader">
	<IxGroupItem text="Group item 1" />
	<IxGroupItem text="Group item 2" />
	<IxGroupItem text="Group item 3" />
</IxGroup>
```

danielleroux:
Will be part of the next release https://github.com/siemens/ix/pull/1257 (after PR is merged)

PR #1225, opened 2024-04-17
fix(core/group): use correct cursors

ridvandmrc:
…d by user

- group-item cursor is default which is not correct

GitHub Issue Number: #1207 

- initially, group item cursor is pointer
- and also, the user can override it because it inherits from host element

## 🏁 Checklist

- [X] 🦮 Accessibility (a11y) features were implemented
- [X] 🗺️ Internationalization (i18n) - no hard coded strings
- [X] 📲 Responsiveness - components handle viewport changes and content overflow gracefully
- [X] 📄 Documentation was reviewed/updated (`pnpm run docs`)
- [X] 📸 Visual regression tests were added/updated and pass ([Guide](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md#visual-regression-testing))

### DIFF
--- a/packages/core/src/components/group-item/group-item.scss
+++ b/packages/core/src/components/group-item/group-item.scss
@@ -20,7 +20,6 @@ $focusBorderWidth: 1px;
   display: flex;
   min-height: 2.25rem;
   height: 2.25rem;
-  cursor: pointer;
 
   > button {
     display: flex;
@@ -33,7 +32,7 @@ $focusBorderWidth: 1px;
     background-color: var(--theme-group-item--background);
     border: 1px solid transparent;
     color: var(--theme-color-std-text);
-    cursor: inherit;
+    cursor: pointer;
 
     padding-left: 2.5rem;
 
@@ -74,6 +73,10 @@ $focusBorderWidth: 1px;
   }
 }
 
+:host(.suppress-selection) {
+  pointer-events: none;
+}
+
 :host(.selected) {
   border-top-width: 0.062rem !important;
   background-color: var(--theme-color-ghost--selected);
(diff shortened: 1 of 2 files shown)


packages/core/src/components/group-item/group-item.scss
```diff
@@ -20,6 +20,7 @@ $focusBorderWidth: 1px;
   display: flex;
   min-height: 2.25rem;
   height: 2.25rem;
+  cursor: pointer;
```
nuke-ellington, line 23: Setting the cursor on the button should suffice

packages/core/src/components/group-item/group-item.scss
```diff
@@ -32,6 +33,7 @@ $focusBorderWidth: 1px;
     background-color: var(--theme-group-item--background);
     border: 1px solid transparent;
     color: var(--theme-color-std-text);
+    cursor: inherit;
```
nuke-ellington, line 36: Only show _pointer_ if _suppressSelection_ is not set!

nuke-ellington (changes requested):

packages/core/src/components/group-item/group-item.scss
ridvandmrc: Yes, but in this way , we give an option to manipulate to change cursor.
and also ionic is following same approach:
https://github.com/ionic-team/ionic-framework/tree/main/core/src/components

nuke-ellington: I get your argument, but we will develop a guide line how to handle cursors in the whole design system. Until that will don't want to change the this.

packages/core/src/components/group-item/group-item.scss
ridvandmrc: yes I will do it

nuke-ellington: Great, can you do it today? :D
We are planning to release this week.

ridvandmrc: of course I will do it today :)

nuke-ellington (dismissed):

nuke-ellington (approved):

PR #1230, opened 2024-04-18
fix(core/event-list): event-list-items not selectabe in chrome

### LINKED ISSUE #1179: Event list sometimes stops responding to mouse hover and clicks when adding elements after a modal is closed
### Prerequisites

### What happened?

In one of our Angular projects, we're using the Event list component as follows. An 'Add' button opens a modal in which a name must be provided. After closing the modal, an event list item using the provided name is added to the event list.

In some instances, after performing the above procedure and after the modal is closed, the event list is not responding to mouse hover and mouse clicks anymore.

Unfortunately, this bug does not happen everytime. Sometimes, it happens only after adding multiple items. Sometimes it doesn't occur at all. The attached video illustrates the problem. When adding items 1 and 2 there is no problem. As seen in the video, the event list is still responding to mouse hover an clicks. After adding item 3 however, the event list does not respond to mouse hover and clicks anymore.

As also seen in the video, after opening a modal again and dismissing by pessing 'Cancel', the event list responds to mouse hover and clicks again. The bug was reproduced in Google Chrome. This is the video:

https://github.com/siemens/ix/assets/164166687/bb4df8d8-3df3-44ac-80d7-a9a589a4f0ac

This Stackblitz has the reproduction project which was used in the video:

https://stackblitz.com/edit/siemens-ix-event-list-bug-repro

This bug seems related to the modal. When adding items to the list without using a modal (e.g. by providing a hardcoded name), this bug does not occur.

Angular

v2.1.2

```shell
https://stackblitz.com/edit/siemens-ix-event-list-bug-repro
```

matthiashader:

GitHub Issue Number: Closes #1179, [IX-1070]

- This bug is pretty flaky, which is why no test case has been written
- When creating an event-list-item through a modal (the modal has an input field and passes the content via the close result to the event-list) - it can sometimes happen that no / or some event-list-items are not selectable anymore. Also the hover state does not work correctly when this error occurs.
- In the initial issue a 'workaround' was discovered, where this bug does not happen if the animate attribute is set to false.
- This issue seems only to occur in chrome, firefox does not have this problem
- .animate is replaced with animejs
- resizing the window does also resolve the issue 

- .animate jQuery was removed, as this could have been the only source of error and replaced with animejs
- it seems that the animation from the modal is conflicting with the jquery animate (eventually has smth to do with the js callstack?)
- event-list-item is rendered and hoverable

## 🏁 Checklist


nuke-ellington (changes requested):
Changes seem fine. During review I stumbled over some odd things that we should address:

- Why is fade out duration 50?
- If a new items gets added it immediately appears in the DOM, then the fade out animation runs followed by fade in. Makes now sense :-)

nuke-ellington (approved):

PR #1280, opened 2024-05-16
feat(vue): add v-model support to ix-select and ix-toggle

tiagogviegas:

IxSelect and IxToggle don't support v-model directive out of the box. Being components that often are used in form control v-model support should exist.

- Updated vueOutputTarget options to enable support for v-model binding in ix-select and ix-toggle.

## 🏁 Checklist

### DIFF
--- a/.changeset/yellow-rabbits-behave.md
+++ b/.changeset/yellow-rabbits-behave.md
@@ -1,0 +1,6 @@
+---
+"@siemens/ix": minor
+"@siemens/ix-vue": minor
+---
+
+feat(vue): add v-model support to ix-select and ix-toggle
--- a/packages/core/stencil.config.ts
+++ b/packages/core/stencil.config.ts
@@ -76,6 +76,18 @@ export const config: Config = {
       includePolyfills: false,
       includeDefineCustomElements: false,
       excludeComponents: ['ix-playground-internal', 'ix-icon'],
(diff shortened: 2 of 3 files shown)


nuke-ellington (commented):
PR looks good but it might look even nicer if we could set up v-model support for all iX components ;-) Not sure how much effort this would be though..

tiagogviegas:
> PR looks good but it might look even nicer if we could set up v-model support for all iX components ;-) Not sure how much effort this would be though..

Do we have a list of which components make sense to add v-model support? In my opinion it would only make sense to add it to components which have a 2-way data binding logic.

nuke-ellington:
> > PR looks good but it might look even nicer if we could set up v-model support for all iX components ;-) Not sure how much effort this would be though..
> 
> Do we have a list of which components make sense to add v-model support? In my opinion it would only make sense to add it to components which have a 2-way data binding logic.

Unfortunately no. One would have to check all components manually.

danielleroux:
We be implemented in https://github.com/siemens/ix/pull/1318

Also found an bug inside stencil vue output target which is patched in PR https://github.com/siemens/ix/pull/1318/files#diff-20e3815ee3dd23189c10abdc94e28f41885810c2e2bea70ff209834f5f9121b8R9-R11

Close this one here

Issue #1076, opened 2024-02-08
ix-select not working when applying data using Angular's ReactiveForms


marcossiemens:
### Prerequisites

- [X] I have not leaked any internal company information.
- [X] I have not posted internal links or code snippets.

### What happened?

When I apply data inside a formGroup using Angular's ReactiveForms like so: `this.form.get('select').setValue(this.data.select.selectId);`, the ix-select doesn't show the item selected.

### What type of frontend frameware are you seeing the problem on?

Angular

### Which version of iX do you use?

2.0.4

### Code to produce this issue.

```shell
Typescript:
this.form.get('select').setValue(this.data.select.selectId);

HTML:
<ix-select id="select" formControlName="select"> <ix-select-item *ngFor="let select of selects" [label]="select.selectName" [value]="select.selectId"> </ix-select-item></ix-select>
```

matthiashader:
Internal ref [IX-912]

jul-lam:
There is no need to wait for the next patch release, since using a string value for the `value` property of `ix-select-item` fixes the issue. In your case changing the `selectId` from `number` to `string` does the job.

**Note**: In future releases the `value` property will be of type `string` instead of `any`.

marcossiemens:
@jul-lam it works as you said. Thanks!

PR #1287, opened 2024-05-21
fix(core/select): rollback and fix object iteration

### LINKED ISSUE #1270: ix-select does not allow looping over array of objects anymore since v2.2.0
### Prerequisites

### What happened?

Since iX v2.2.0 the value property of ix-select-item must be a string. That is enforced by this PR: https://github.com/siemens/ix/pull/1159

That change has broken our implementation, as we're using ix-select and ix-select-item to display a select box of object from an array.
Due to the changes in PR https://github.com/siemens/ix/pull/1159, since ix v2.2.0, the selected item is not displayed anymore in the select box in our implementation. I've created a Stackblitz to illustrate the problem: https://stackblitz.com/edit/siemens-ix-select-issues. In the Stackblitz:

- Example 1 illustrates how the official documentation expects the user to use ix-select
- Example 2 illustrates how we're using ix-select. Note that when selecting an item in the select box, the selected item is not displayed in the select box (that used to work in iX v2.1.2, but does not work anymore since iX v2.2.0). The selected item is however still set correctly in Typescript.
- Example 3 converts our implementation to use strings instead of objects as the value of ix-select-item. That implementation is not optimal though because, to get the actually selected item from the array, some additional code must be written in the valueChange handler to fetch the real object from the array, based on the string that is passed in the valueChange CustomEvent. 
- Example 4 illustrates how a standard html select works directly with objects. Which is how ix-select used to work and ideally should work.

Example 3 makes ix-select work with objects, given that the objects have a property of type string that can be used as ix-select-item value. However, extra code must be added to the valueChange handler to get the object from the original array.
Is ix-select supposed to behave this way? Is it expected from users that such extra code must be written in the valueChange handler to get the selected object?

Angular

v2.2.0

```shell
https://stackblitz.com/edit/siemens-ix-select-issues
```

matthiashader:
- #1076 adapts the type conversion of the selected item to string
- object iteration does not work anymore

GitHub Issue Number: Closes #1270 , [IX-1235]

- rollbacked changes
- added warning if non string values are used inside the select
- implemented type conversion on mismatch (needed?)

## 🏁 Checklist


nuke-ellington (dismissed):

danielleroux (approved):

packages/core/src/components/select/select.tsx
```diff
@@ -350,6 +349,12 @@ export class Select {
       this.value = this.selectedIndices;
     }
 
+    if (typeof this.items[0]?.value !== 'string') {
+      console.warn(
+        'Non string data types not supported and and will be removed in 3.0'
+      );
+    }
+
```
danielleroux, line 357: Could be a potential undefined access if no items are provided. Maybe a better place to show this warning will be inside the forEach above.

danielleroux (changes requested):

danielleroux (approved):

Issue #1202, opened 2024-04-09
[Firefox] TabControl is not displayed correctly


alex50105:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [X] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

When using Firefox, the Tab does not display correctly. It does not show the underlines, it makes it impossible to see which tab is selected.

The following screenshot are from the documentation page: https://ix.siemens.io/docs/controls/tabs/
But we also observed in our project using v2.1.0.

Chrome/Edge:
![image](https://github.com/siemens/ix/assets/166496797/18188698-b15f-4c28-b5af-b3f7074719a1)
![image](https://github.com/siemens/ix/assets/166496797/5e2e591a-f101-4792-bd38-341cd2d0acfb)

Firefox (124.0.2):
![image](https://github.com/siemens/ix/assets/166496797/d2f10b51-ce90-4764-9b86-af66f2dfbd9f)
![image](https://github.com/siemens/ix/assets/166496797/20fd5f9d-8a32-4519-8297-9a99e9e20508)

### What type of frontend framework are you seeing the problem on?

Angular

### Which version of iX do you use?

v2.1.0

### Code to produce this issue.

```shell
the samples on the documentation page https://ix.siemens.io/docs/controls/tabs/
```

alex50105:
this was fixed with the release of 2.3.0, thanks

PR #1313, opened 2024-06-06
fix(core/scrollbar): fix firefox appearance (scrollbar & input & tabs)

matthiashader:
- Input, spefically the pagination has the increment / decrement buttons shown
- Default Scrollbar applied in Firefox

GitHub Issue Number: #N/A, [IX-1126], [IX-1127], [IX-1207]
Note: Viva Engage Firefox issue has also been adapted

- Increment / Decrement buttons are hidden
- Unfortunately Firefox does not offer that big of a support of scrollbar styling, so only available classes have been used (see https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_scrollbars_styling)
- Note: There is a difference between the default Firefox Scrollbar in Windows 10 and Windows 11!
- No Tests, since we only test in Chrome

## 🏁 Checklist

### DIFF
--- a/packages/core/src/components/tab-item/tab-item.scss
+++ b/packages/core/src/components/tab-item/tab-item.scss
@@ -25,7 +25,7 @@
   background-color: var(--theme-tab--background);
   color: var(--theme-tab--color);
 
-  ::after {
+  &::after {
     content: '';
     position: absolute;
     background-color: var(--theme-tab-indicator--background);
@@ -110,7 +110,6 @@
       border-color: var(--theme-tab-pill--border-color--selected);
 
       &::after {
-        display: none;
       }
     }
 
@@ -119,7 +118,6 @@
     }
 
     &::after {
-      display: none;
     }
   }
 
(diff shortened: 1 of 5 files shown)


packages/core/src/components/tab-item/tab-item.scss
```diff
@@ -110,7 +110,6 @@
       border-color: var(--theme-tab-pill--border-color--selected);
 
       &::after {
```
nuke-ellington, line 112: Empty selector

packages/core/src/components/tab-item/tab-item.scss
```diff
@@ -119,7 +118,6 @@
     }
 
     &::after {
```
nuke-ellington, line 120: Empty selector

nuke-ellington (changes requested):

packages/core/src/components/tab-item/tab-item.scss
matthiashader: Removed, seems not to be needed

packages/core/src/components/tab-item/tab-item.scss
matthiashader: Removed, seems not to be needed

nuke-ellington (approved):

danielleroux:
Closes #1202

PR #1397, opened 2024-07-22
doc(demo-app): add text

bettiifr:
Demo App text


kathrinschalber (commented):
Not bad at all! As an idea, maybe you could sync on this with Jennie, she might have inspo for this too

PR #1432, opened 2024-08-14
docs/buttons: edit styleguide

bettiifr:
The styleguide for the buttons is improved here

### DIFF
--- a/packages/documentation/docs/controls/buttons/_button_styleguide.md
+++ b/packages/documentation/docs/controls/buttons/_button_styleguide.md
@@ -2,24 +2,22 @@ Buttons initiate actions, apply actions to selected objects and activate/deactiv
 
 ![Overview](https://www.figma.com/design/wEptRgAezDU1z80Cn3eZ0o/iX-Pattern-Illustrations?type=design&node-id=1097-5037&mode=design&t=KAxDgJoFX436Uk0b-11)
 
-1. Button label
-2. Button icon
+1. Label
+2. Icon
 
 ## Variants
 
-- **Primary button:** In our applications, we most often use the primary button variant.
-- **Secondary button:** The secondary button variant has a grayscale appearance to distinguish it from the primary button.
-- **Danger button:** We use the danger button variant to emphasize particularly dangerous, destructive or critical actions that cannot easily be undone. Consider this especially in confirmation dialogs. 
+- **Primary:** The primary button variant has a solid background in the primary color. We typically use primary buttons for the most important or most-likely next action within the user interface. 
+- **Secondary:** The secondary button variant has a transparent background in the secondary color. We typically use secondary buttons for standard actions that need to be easily recognizable or for actions supporting the default (primary) action. 
+- **Danger:** The danger button variant has a red background and is used for destructive or critical actions like "delete" or "remove", especially in confirmation dialogs. We typically use the danger button for actions that are irreversible or have a significant impact on the user’s data or the application state.  
+- **Ghost :** The ghost button variant has a transparent background and a border in the primary color. We typically use ghost buttons for actions that are not part of the core user journey but serve specialized or conditional purposes. 
+- **Outline**: All Button variants have also an outline variant. The outline variant has a transparent background and a border in the corresponding color. This variant is used when the button needs to be more subtle or when it is placed on a dark background. 
 
 **Note:** Please be aware that the terms *primary* and *secondary* are not used in accordance to common UX terminology for primary, secondary and ghost button.
 
 ## Options
-- **Default:** Use for the most important or most-likely next action within the user interface. These are typically actions that advance the user through a process, such as "Submit", "Save" or "Add". We use these sparingly and recommend only one per layout. These are typically called *primary* buttons in UX. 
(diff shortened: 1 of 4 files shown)


packages/documentation/docs/controls/buttons/_button_styleguide.md
```diff
@@ -2,57 +2,53 @@ Buttons initiate actions, apply actions to selected objects and activate/deactiv
...
+1. Label
+2. Icon
 
 ## Variants
 
-- **Primary button:** In our applications, we most often use the primary button variant.
-- **Secondary button:** The secondary button variant has a grayscale appearance to distinguish it from the primary button.
-- **Danger button:** We use the danger button variant to emphasize particularly dangerous, destructive or critical actions that cannot easily be undone. Consider this especially in confirmation dialogs. 
+- **Primary:** Has a solid background in the primary color. We typically use primary buttons for the most important or most-likely next action within the UI. 
+- **Secondary:** Has a transparent background in the secondary color. We typically use secondary buttons for standard actions that need to be easily recognizable or for actions supporting the default (primary) action.
```
flxlst09, line 11: Description is the same as for the Outline option ("for standard actions actions... supporting the primary action". In Figma the Secondary button is described as "Use for special usecases when other button types catch too much attention". The whole primary/secondary topic is highly confusing, we should spend a paragraph and an illustration on that.

packages/documentation/docs/controls/buttons/_button_styleguide.md
```diff
@@ -2,57 +2,53 @@ Buttons initiate actions, apply actions to selected objects and activate/deactiv
...
-- **Icon:** Icons can also be displayed with button labels.
-- **Disabled:** Buttons can be disabled (see also button states).
-- **Loading:** A loading spinner is displayed on the button. The spinner replaces an icon when available.
-- **Type:** A submit button is available. Submit buttons are typically used in forms and trigger a submit event. Apply this type to make a submit button more accessible.
+
+- **Label**: The label is the text displayed on the button. We typically use short, descriptive labels that clearly communicate the action the button triggers.
+- **Icon:** An icon can be displayed on the button. We typically use icons to support the label and make the button more easy to discover by the user. The icon should be widely known for representing the action or function among your users.
+- **Type:** Use a plain button for actions that do not involve form submission, such as triggering dialogs or performing navigation. Otherwise, use a submit button to send user input from a form to a server.
```
flxlst09, line 20: Would turn it around, starting with the Submit type. The term "plain" button is not reflected in the property name, there, can we avoid it? Btw we should rename the Figma property, there "type" defines primary vs secondary.

Suggestion:
When sending user input from a form to a server, use the type "submit". For all other actions, use the default type "button".

packages/documentation/docs/controls/buttons/_button_styleguide.md
```diff
@@ -2,57 +2,53 @@ Buttons initiate actions, apply actions to selected objects and activate/deactiv
...
-- **Ellipsis (…):** Ellipsis can be used to indicate that an action requires further input or choice from the user, such as "Save as…" which opens a further list of file types to choose from. Ellipses is typically not used for actions with a subsequent confirmation dialog.
-- **Alignment:** Buttons can be left-justified or right-justified or fully span a container’s width.
-- **Cluster buttons:** Buttons can be clustered in groups based on their relationship. A cluster can contain a mixture of buttons, e.g. a combination of a primary button with two primary ghost buttons. We recommend a minimum distance of `0.5rem` between adjacent buttons, and we typically cluster buttons if actions/functions are related but not excluding each other. Otherwise, consider using a `button group`. 
-- **Button width:** Button width depends on its content. In addition, buttons have a default minimum width of 5rem to lay out common combinations such as "OK" and "Cancel" more harmoniously with equal widths. The minimum width is customizable to accommodate other combinations.
+- **Alignment:** Buttons can be left- or right-justified or fully span a container’s width.
```
flxlst09, line 28: better right-aligned? would use the term 'justified' only for typography, but not sure.

packages/documentation/docs/controls/buttons/_button_styleguide.md
```diff
@@ -2,57 +2,53 @@ Buttons initiate actions, apply actions to selected objects and activate/deactiv
...
+- **Cluster buttons:** Cluster buttons in groups with related functions. A cluster might include various types of buttons, e.g. a primary and  secondary ghost button. We recommend a gap of `0.5rem` between buttons.
```
flxlst09, line 29: would give an example that shows the intended use of primary and primary outline ("primary and secondary in UX context").

flxlst09 (commented):
Minor things regarding vocabulary. Biggest topic is whether we can provide a better explanation in regard of the Primary vs Secondary confusion.

packages/documentation/docs/controls/buttons/_button_styleguide.md
kathrinschalber: that's much better!

packages/documentation/docs/controls/buttons/_button_styleguide.md
kathrinschalber: true - changed ✔️

PR #1457, opened 2024-09-04
fix(core/pane-layout): remove absolute positioning

### LINKED ISSUE #1305: Pane component does not respect parent size
### Prerequisites

### What happened?

The Pane component does not respect its parent size. When putting ix-pane in a parent div, ix-pane bleeds out of the parent div. This is due to **position: absolute** which is set on the **:host** of the pane component. When overriding positioning and setting it to **unset**, the sizing occurs as expected and the Pane respects its parent size.

This StackBlitz illustrates the issue: https://stackblitz.com/edit/ix-pane-positioning

Angular

v2.1.2

```shell
https://stackblitz.com/edit/ix-pane-positioning
```

matthiashader:
- pane-layout is positioned absolute, despite being inside a  div with height/width restriction it takes the whole screen
- workaround is to overwrite the positioned: absolute css prop

GitHub Issue Number: Closes #1305, [IX-1322]

- position of the pane-layout is relative
- default it takes 100% height / 100% width

## 🏁 Checklist


nuke-ellington (commented):
This is a breaking change IMO.

nuke-ellington (changes requested):
Maybe we could come up with a more descriptive name for the added test case. I am afraid that six months from now no one will know what was tested. Maybe instead of "relative" something like "responsive" could work?

packages/core/src/tests/panes/panes.e2e.ts
```diff
@@ -142,6 +142,13 @@ test.describe('pane', () => {
     await expect(page).toHaveScreenshot();
   });
 
+  regressionTest('layout, inline, collapsed, relative', async ({ page }) => {
```
nuke-ellington, line 145: You need to regenerate the snapshots
```suggestion
  regressionTest('layout, inline, collapsed, responsive', async ({ page }) => {
```

nuke-ellington (approved):

Issue #1228, opened 2024-04-17
Tooltip in modal popup doesn't use available space


Stefei:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [X] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

Tooltips inside modals don't use all available space. Below is the code to include in the modal to reproduce the issue

For the screenshot we used a modal with size 840:

![grafik](https://github.com/siemens/ix/assets/82656629/bc976170-fa71-4a15-a655-c542d40323d2)

### What type of frontend framework are you seeing the problem on?

Angular, JavaScript

### Which version of iX do you use?

v2.1.0

### Code to produce this issue.

```shell
<div style="display: flex; margin-top:600px; margin-bottom:600px">
    <input checked id="checkbox_1_1" name="group_1" type="checkbox" />
    <label for="checkbox_1_1"> Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</label>
    <div id="test">
      <ix-icon size="24" name="name" name="info"></ix-icon>
    </div>
    <ix-tooltip style="align-self: flex-start;" for="#test">Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.
    </ix-tooltip>
  </div>
```

PR #1462, opened 2024-09-06
fix(core/tooltip): cutoff near viewport edges

### LINKED ISSUE #740: Tooltip visibility
### What happened?

In certain cases the tooltip does not shown correctly, because if near the side of the view it overflows and only half of it is visible. In addition if there is an 'icon' or 'pill' near the tooltip, the tooltip is behind and not on top, so also not clearly visible. (this can be fixed if you are not using visibility attribute but toggle display, to not be rendered in the DOM when not hovered in my opinion.) 
image
image

Angular

v1.6.3

_No response_

matthiashader:
- If a tooltip gets to near  to the view port  edge the tooltip will get cutoff
- It seems  that the PR from [IX-901] leads to an issue, where the border-radius will not get  adapted

GitHub Issue Number: Closes #740,  [IX-849]

-  Tooltip will automatically gets placed (if there isn't enough space) in following order:  ['right', 'left', 'top', 'bottom']
- Border radius gets inherited

## Note
- computePosition also return  a  top property, when 'fallbackAxisSideDirection' is set, this result in missplacement of the arrow
- Potential source: 
image
As the postion is set during computing and the middlewareData is deepmerged, this can result in the unwanted  top  property -> As postion = 'top' doesn't need  the  property, it has been removed

## 🏁 Checklist

### DIFF
--- a/packages/core/src/components/tooltip/tooltip.scss
+++ b/packages/core/src/components/tooltip/tooltip.scss
@@ -1,8 +1,8 @@
 :host {
   display: inline-block;
   position: fixed;
-  left: 0px;
-  top: 0px;
+  left: 0;
+  top: 0;
   z-index: var(--theme-z-index-tooltip);
 
   max-width: 18.25rem;
@@ -32,7 +32,7 @@
     padding: 0.375rem 0.75rem 0.375rem 0.875rem;
 
     box-shadow: var(--theme-shadow-4);
-
+    border-radius: inherit;
   }
 }
 
--- a/packages/core/src/components/tooltip/tooltip.tsx
+++ b/packages/core/src/components/tooltip/tooltip.tsx
@@ -139,7 +139,6 @@ export class Tooltip implements IxOverlayComponent {
     if (placement.startsWith('top')) {
       return {
         left: numberToPixel(x),
-        top: numberToPixel(y),
       };
     }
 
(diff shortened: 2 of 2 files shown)


danielleroux (changes requested):
Fix cause a flicker effect using the steps: https://github.com/siemens/ix/issues/1228

Screen Recording 2024-09-09 at 15 18 48


Here a runnable example code: 

```tsx
export class PlaygroundInternal {
  componentDidLoad() {
    const content = document.createElement('DIV');
    content.style.display = 'flex';
    content.style.marginTop = '600px';
    content.style.marginBottom = '600px';
    content.innerHTML = `
      <input checked id="checkbox_1_1" name="group_1" type="checkbox" />
      <label for="checkbox_1_1"> Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.</label>
      <div id="test" style="display: block; width: 5rem; height: 5rem; background: red">
      </div>
      <ix-tooltip style="align-self: flex-start;" for="#test">Lorem ipsum dolor sit amet, consetetur sadipscing elitr, sed diam nonumy eirmod tempor invidunt ut labore et dolore magna aliquyam erat, sed diam voluptua. At vero eos et accusam et justo duo dolores et ea rebum. Stet clita kasd gubergren, no sea takimata sanctus est Lorem ipsum dolor sit amet.
      </ix-tooltip>
    `;

    showModal({
      content: content,
    });
  }

  render() {
    return <Host></Host>;
  }
}

```

danielleroux (approved):

PR #1557, opened 2024-11-11
fix(core/select): check if value is defined, before updating selection

matthiashader:
Addition to #1548
- select in angular logs an error to the console when initializing the select-items in async 
- last element has always an undefined value
- angular seems to initialize components differently than react (possibly due to the vdom)

GitHub Issue Number: N/A, [IX-1918]

- before the selection is updated, it is checked if the value is  defined
- valueChange emited and selection updated
- updateSelection still needed when select properties change!
- removed "get dropdownItem()", method is never called - only appearance in select.tsx does not call this methods 
- only needed for the addItem, which is not a select-item (it is a dropdown-item)
- label / value changes only occur after the components has been initalised -> fewer events are thrown, also the slotChange takes care about added / deleted slots
## 🏁 Checklist

### DIFF
--- a/packages/core/src/components/select-item/select-item.tsx
+++ b/packages/core/src/components/select-item/select-item.tsx
@@ -22,14 +22,13 @@ import {
   IxSelectItemLabelChangeEvent,
   IxSelectItemValueChangeEvent,
 } from './events';
-import { DropdownItemWrapper } from '../dropdown/dropdown-controller';
 
 @Component({
   tag: 'ix-select-item',
   styleUrl: 'select-item.scss',
   shadow: true,
 })
-export class SelectItem implements DropdownItemWrapper {
+export class SelectItem {
   @Element() hostElement!: HTMLIxSelectItemElement;
 
   /**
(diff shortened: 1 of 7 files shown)


packages/core/src/components/select-item/select-item.tsx
```diff
@@ -55,12 +58,14 @@ export class SelectItem implements DropdownItemWrapper {
...
-  @Event() itemClick: EventEmitter<string>;
+  @Event() itemClick!: EventEmitter<string>;
+
+  private componentLoaded = false;
 
   /** @internal */
   @Method()
   async getDropdownItemElement(): Promise<HTMLIxDropdownItemElement> {
-    return this.dropdownItem;
+    return this.dropdownItem as HTMLIxDropdownItemElement;
```
nuke-ellington, line 68: Possibly better solution?
```suggestion
    return this.dropdownItem!;
```

packages/core/src/components/select/select.tsx
```diff
@@ -369,19 +368,14 @@ export class Select {
...
     event.stopImmediatePropagation();
     this.updateSelection();
   }
 
-  disconnectedCallback() {
-    if (this.itemMutationObserver) {
-      this.itemMutationObserver.disconnect();
-    }
-  }
-
```
nuke-ellington, line 384: Why this change?

nuke-ellington (changes requested):

nuke-ellington (approved):

PR #1651, opened 2025-01-16
fix(core/radio): remove :disable selector to render disable correct

ridvandmrc:
## 💡 What is the current behavior?

I was not working properly if the disabled attribute like disabled='false'
GitHub Issue Number: #1628

## 🆕 What is the new behavior?

-
-

## 🏁 Checklist

A pull request can only be merged if all of these conditions are met (where applicable):

- [X] 🦮 Accessibility (a11y) features were implemented
- [X] 🗺️ Internationalization (i18n) - no hard coded strings
- [X] 📲 Responsiveness - components handle viewport changes and content overflow gracefully
- [X] 📄 Documentation was reviewed/updated (`pnpm run docs`)
- [X] 🧪 Unit tests were added/updated and pass (`pnpm test`)
- [ ] 📸 Visual regression tests were added/updated and pass ([Guide](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md#visual-regression-testing))
- [X] 🧐 Static code analysis passes (`pnpm lint`)
- [X] 🏗️ Successful compilation (`pnpm build`, changes pushed)

PR #1652, opened 2025-01-17
fix(core/checkbox): remove :disable selector to render disable correct

danielleroux:

Followup of #1651.

Removing `:disabled` selector ensures that the host component is also valid if used with react.

e.g

```
<ix-checkbox disabled={undefined} />
```

## 🏁 Checklist


packages/core/src/components/checkbox/tests/checkbox.ct.ts
```diff
@@ -50,6 +50,33 @@ test(`disabled`, async ({ mount, page }) => {
...
+    }
+  );
+
+  await checkboxElement.click();
+  await checkedChange$;
+
+  await expect(checkboxElement).not.toHaveClass(/disabled/);
+  await expect(nativeInput).not.toBeDisabled();
+
+  const disableLabelColor = 'rgba(245, 252, 255, 0.93)';
```
AndreasBerliner, line 76: I would suggest renaming this const because this is not the text-color for the disabled state, but for the enabled state.

packages/core/src/components/checkbox/tests/checkbox.ct.ts
```diff
@@ -50,6 +50,33 @@ test(`disabled`, async ({ mount, page }) => {
...
+        element.addEventListener('checkedChange', () => resolve());
+      });
+    }
+  );
+
+  await checkboxElement.click();
+  await checkedChange$;
+
+  await expect(checkboxElement).not.toHaveClass(/disabled/);
+  await expect(nativeInput).not.toBeDisabled();
```
AndreasBerliner, line 74: I noticed that the attribute: 'aria-disabled' has the value: 'undefined' in this case. I would expect it not to be present (like the disabled class).

`expect(await checkboxElement.getAttribute('aria-disabled')).toBeNull();`

packages/core/src/components/checkbox/tests/checkbox.ct.ts
danielleroux: Suggestion?

AndreasBerliner: Maybe 'checkboxLabelColor' ?

AndreasBerliner (approved):

PR #1658, opened 2025-01-20
fix(core/pill): fix pill component layout

AndreasBerliner:

- The padding on the left is not correct for the pill component when an icon is set.
- The padding on the right is off by 1px for the pill component.
- The documentation lacks an example with only an icon and no text.
- The spacing is incorrect when only an icon and no text is used.

- The padding on the left is is now applied correctly when an icon is set on the pill component.
- The padding on the right is now applied correctly.
- The spacing is now correct, when only an icon and no text is used.

## 🏁 Checklist


packages/core/src/components/pill/pill.scss
```diff
@@ -22,12 +23,16 @@
 }
 
 @include chip-container(false);
+@include pill-container();
```
danielleroux, line 26: Whats the reason behind the additional mixin here? There is no other files which depending on pill styles.
Now you have to jump into an additional mixin to check the styling without any benefits from moving it into a mixin

packages/core/src/components/pill/pill.tsx
```diff
@@ -63,6 +63,13 @@ export class Pill {
    */
   @Prop() alignLeft = false;
 
+  private applySpacing() {
+    if (!this.hostElement.textContent) {
+      return false;
+    }
+    return this.hostElement.textContent.length >= 1;
```
danielleroux, line 70:
```suggestion
    return this.hostElement.textContent.length > 0;
```

.changeset/cold-shirts-vanish.md
```diff
@@ -0,0 +1,6 @@
+---
+'@siemens/ix': patch
+---
+
+fix `ix-pill` layout/spacing, adapt tests and add examples to docs
```
danielleroux, line 5:
```suggestion
Adjust `ix-pill` spacing between text + icon and icon only
```

Information about testing und examples is not that relevant for an changelog entry

packages/core/src/components/pill/pill.tsx
```diff
@@ -63,6 +63,13 @@ export class Pill {
...
+  }
+
```
danielleroux, line 72: Looks like dynamic removing text does not remove the gap.
```
import { test } from '@utils/test';
import { expect } from '@playwright/test';

test.describe('Pill', () => {
  test('dynamic gap behavior', async ({ mount, page }) => {
    await mount('<ix-pill icon="star">Dynamic gap</ix-pill>');
    const pill = page.locator('ix-pill');
    const innerContainer = pill.locator('.container');

    await expect(innerContainer).toHaveClass(/with-gap/);

    await pill.evaluate((el: HTMLIxPillElement) => {
      el.innerText = '';
    });

    await expect(innerContainer).not.toHaveClass(/with-gap/);
  });
});

```

danielleroux (changes requested):

danielleroux (approved):

Issue #1570, opened 2024-11-20
ix-select element grows in width when entering large text


yetibrain:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [X] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

we got some difficulties regarding the usage of the ix-select element as well as ix-select-item elements. We have set the attribute editable to true on the ix-select element. When having a large selection text, it leads to a bigger dropdown. When editing and entering many characters, it happens, that the element grows in width, even wider than the screen size! Using width and/or max-width css styles didn't change the behavior.

### What type of frontend framework are you seeing the problem on?

Angular

### Which version of iX do you use?

latest

### Code to produce this issue.

```shell
See https://stackblitz.com/edit/bmfent-pbpvcx?file=src%2Fselect.ts,package.json
```

network-interface-controller:
Hi @AndreasBerliner, what's the current state of this issue? Does the linked MR fix the described problem?

PR #1669, opened 2025-01-28
fix(core/select): prevent list items to expand beyond screen width and add dropdown width properties

AndreasBerliner:

- For drop-down items with longer texts, the item width can be larger than the available space.
- There is no way to set the width of the drop-down list.
- There is no way to set the maximum width of the drop-down list.

GitHub Issue Number: #1570 

- The default maximum width of drop-down list was set to 100%, which prevents the item width to be larger than the available space.
- A property to set the width of the drop-down list was introduced.
- A property to set the maximum width of the drop-down list was introduced.

## 🏁 Checklist


danielleroux:
@AndreasBerliner 
image

1. Start storybook
2. Open http://localhost:6006/?path=/story/example-select--editable-with-dropdown-width&args=dropdownMaxWidth:1000&globals=viewport:mobile1


Looks like select is still cut off on right side.

packages/core/src/components/select/select.tsx
```diff
@@ -960,6 +976,8 @@ export class Select implements IxInputFieldComponent<string | string[]> {
 
             return {
               minWidth: `${minWidth}px`,
+              width: `${this.dropdownWidth}rem`,
+              maxWidth: `${this.dropdownMaxWidth}rem`,
```
danielleroux, line 980: If you allow here only number values and add a suffix `rem` to it prevents user from building response bahavior e.g 50vw or 50%

packages/core/src/components/select/select.scss
```diff
@@ -160,6 +160,10 @@
     }
   }
 
+  ix-dropdown {
+    max-width: 100%;
```
danielleroux, line 164: `Defines the max-width as a percentage of the containing block's width` this means the ix-select host will be the 100%. 
Or did I see this wrong?

danielleroux (changes requested):

matthiashader:
On editable mode, this behavior looks a little strange, eventually we could consider adapting the width also based on the content? 
image

packages/core/src/components/select/select.tsx
```diff
@@ -179,6 +179,22 @@ export class Select implements IxInputFieldComponent<string | string[]> {
    */
   @Prop() hideListHeader = false;
 
+  /**
+   * Optional paramater to set the width of the dropdown element in rem.
```
matthiashader, line 183:
```suggestion
   * Optional parameter to set the width of the dropdown element in rem.
```

matthiashader (changes requested):

packages/core/src/components/select/select.scss
danielleroux: Correct its not the case here

packages/core/src/components/select/select.tsx
```diff
@@ -976,9 +976,13 @@ export class Select implements IxInputFieldComponent<string | string[]> {
...
-              width: this.dropdownWidth,
-              maxWidth: this.dropdownMaxWidth,
-            };
+              ...(this.dropdownWidth && {
+                width: `min(${this.dropdownWidth}, 100%)`,
+              }),
+              ...(this.dropdownMaxWidth && {
+                'max-width': `min(${this.dropdownMaxWidth}, 100%)`,
+              }),
+            } as CSSStyleDeclaration;
```
danielleroux, line 985: 1. Casting not necessary here
2. Just a idea to make the code more readable:
```
            const styleOverwrite: Partial<CSSStyleDeclaration> = {};

            const minWidth = this.hostElement.shadowRoot
              ?.querySelector('.select')
              ?.getBoundingClientRect().width;

            if (minWidth) {
              styleOverwrite.minWidth = `${minWidth}px`;
            }

            if (this.dropdownWidth) {
              styleOverwrite.width = `min(${this.dropdownWidth}, 100%)`;
            }

            if (this.dropdownMaxWidth) {
              styleOverwrite.maxWidth = `min(${this.dropdownMaxWidth}, 100%)`;
            }

            return styleOverwrite;
```

danielleroux (changes requested):

danielleroux (approved):

matthiashader (approved):

Issue #1656, opened 2025-01-20
Ix chip active not working


JakubRigoci:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

Settings propery active to false in ix chip does not do what it says. It stays interactable. Any way to resolve it?

### What type of frontend framework are you seeing the problem on?

Angular

### Which version of iX do you use?

v2.6.0

### Code to produce this issue.

```shell
https://stackblitz.com/edit/3pafz6cw-w8xwbbkf?file=src%2Fchip.ts
```

danielleroux:
Closed by #1683

PR #1683, opened 2025-02-07
fix(core/chip): applied inactive state to host

ridvandmrc:

Settings propery active to false in ix chip does not do what it says. It stays interactable. 

GitHub Issue Number: #1656

- previous implementation was wrong I just moved the inactive class from container to host
- 

## 🏁 Checklist

- [X] 🦮 Accessibility (a11y) features were implemented
- [X] 🗺️ Internationalization (i18n) - no hard coded strings
- [X] 📲 Responsiveness - components handle viewport changes and content overflow gracefully
- [X] 📄 Documentation was reviewed/updated (`pnpm run docs`)


danielleroux:
> I might be change scss file but this usage seems better for me. If you want to update scss file I mean move the style from host to container class please let me know

Its fine how you did it. We consider a general rework of host component selectors to reduce conflicts with other css frameworks, but its not yet finalized.

AndreasBerliner:
Hi @ridvandmrc and thank you for your contribution.
With your approach not all css styles are applied, since the 'inactive' CSS class is expected on the container element. One solution would be to add this class to both the host and the container element.

ridvandmrc:
> Hi @ridvandmrc and thank you for your contribution. With your approach not all css styles are applied, since the 'inactive' CSS class is expected on the container element. One solution would be to add this class to both the host and the container element.

ridvandmrc:
> Hi @ridvandmrc and thank you for your contribution. With your approach not all css styles are applied, since the 'inactive' CSS class is expected on the container element. One solution would be to add this class to both the host and the container element.

I see you point, I added 'inactive in container.

AndreasBerliner (approved):

AndreasBerliner (approved):

PR #1695, opened 2025-02-17
fix(core/message-bar): hide entire `ix-message-bar` and turn off user-interactions

### LINKED ISSUE #1608: Message bar is only invisible after dismissing but still usable
### Prerequisites

### What happened?

We implemented the error message bar with the button to dimiss it. But it happens, that it is invisible, but if you keep clicking on the spot, where the dismiss button is, it appears again and again for a short time after getting invisible again. 

I also looked in the chrome dev tools and it seems, that only the opacity of the message bar is set to 0%. But dismiss would mean for me, that the message bar is removed from the dom, or the bar is not set to opacity 0% so the developer sees, that it is NOT away and you have to implement the event yourself.

Angular

v2.5.0

```shell
import { Component } from '@angular/core';

@Component({
  selector: 'app-example',
  template: `
    <div class="message-bar">
      <ix-message-bar>Message text</ix-message-bar>
      <ix-message-bar type="warning">Message text</ix-message-bar>
      <ix-message-bar type="danger">
        <div class="d-flex align-items-center justify-content-between">
          Message text
          <ix-button>Action</ix-button>
        </div>
      </ix-message-bar>
    </div>
  `,
  styleUrls: ['./message-bar.css'],
})
export default class MessageBar {}
```

matthiashader:
- Examples of DOM removal already done in the last PR
- Button interactions still possible
- d-none applied but not implemented

GitHub Issue Number: Closes #1608, [IX-2201]

- implemented d-none
- Buttons not reachable through mouse or keyboard interactions
- also NVDA screen reader was tried -> buttons are not interactable

## 🏁 Checklist

### DIFF
--- a/.changeset/silly-pandas-peel.md
+++ b/.changeset/silly-pandas-peel.md
@@ -1,5 +1,0 @@
----
-'@siemens/ix': major
----
-
-Hide the entiner `ix-messsage-bar` and disable all user interactions with it
--- a/.changeset/silly-pandas-peel.md
+++ b/.changeset/silly-pandas-peel.md
@@ -1,0 +1,5 @@
+---
+'@siemens/ix': major
+---
+
+Hide the entiner `ix-messsage-bar` and disable all user interactions with it
--- a/packages/core/src/components/message-bar/message-bar.scss
+++ b/packages/core/src/components/message-bar/message-bar.scss
@@ -59,7 +59,7 @@
     margin-top: $tiny-space;
   }
 
-  .d-none {
+  .message-bar-hidden {
     display: none;
   }
 }
--- a/packages/core/src/components/message-bar/message-bar.scss
+++ b/packages/core/src/components/message-bar/message-bar.scss
@@ -58,4 +58,8 @@
   ix-icon {
     margin-top: $tiny-space;
   }
+
+  .d-none {
+    display: none;
+  }
 }
--- a/packages/core/src/components/message-bar/test/message-bar.ct.ts
+++ b/packages/core/src/components/message-bar/test/message-bar.ct.ts
@@ -32,7 +32,7 @@ test.describe('ix-message-bar', () => {
     await expect(messageBar).not.toBeVisible();
   });
 
-  test('adds d-none class after dismissing', async ({ page }) => {
+  test('adds message-bar-hidden class after dismissing', async ({ page }) => {
     const messageBar = page.locator('ix-message-bar');
     const closeButton = messageBar.locator('[data-testid="close-btn"]');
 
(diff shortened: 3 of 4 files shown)


packages/core/src/components/message-bar/test/message-bar.ct.ts
```diff
@@ -0,0 +1,39 @@
...
+          element.addEventListener('closeAnimationCompleted', (e) =>
+            resolve(e)
+          );
+        });
+      }),
+      closeButton.click(),
+    ]);
+
+    expect(event).toBeTruthy();
+  });
```
danielleroux, line 27:
```suggestion
  test('emits closeAnimationCompleted event', async ({ page }) => {
    const messageBar = page.locator('ix-message-bar');
    const closeButton = messageBar.locator('[data-testid="close-btn"]');
   
    const onCloseAnimationCompleted = messageBar.evaluate((...) => {
           return new Promise<void>(resolve => {
                element.addEventListener('closeAnim...', resolve);
           })
    })
    await closeButton.click();
    await onCloseAnimationCompleted;

    expect(messageBar).not.toBeVisible(); // Maybe code have to be adapter here
  });
```

packages/core/src/components/message-bar/message-bar.scss
```diff
@@ -58,4 +58,8 @@
   ix-icon {
     margin-top: $tiny-space;
   }
+
+  .d-none {
+    display: none;
+  }
```
danielleroux, line 64: We should find a better name. `d-none` was a util class from bootstrap.

danielleroux (changes requested):

.changeset/silly-pandas-peel.md
danielleroux, line 1: This change is no `major` change, and overall this is a follow up of an existing changeset. Please edit existing changeset if you need to adapt the changelog message.

danielleroux (changes requested):

.changeset/silly-pandas-peel.md
matthiashader: major was not indentended - but nevertheless has been removed

packages/core/src/components/message-bar/test/message-bar.ct.ts
```diff
@@ -0,0 +1,45 @@
...
+    const messageBar = page.locator('ix-message-bar');
+    const closeButton = messageBar.locator('[data-testid="close-btn"]');
+
+    await closeButton.click();
+    await page.waitForTimeout(300);
+
+    const messageContainer = messageBar.locator('.message-container');
+    await expect(messageContainer).toHaveClass(/message-bar-hidden/);
+  });
+});
```
danielleroux, line 45: What is the different between the two test cases? The first test case test the event and the visibility.

danielleroux (approved):

Issue #1684, opened 2025-02-10
event-list not working with dynamic content and custom height


FabLax:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

When dynamically changing the content of the ix-event-list (e.g. using ngFor or  @ for), newly added event-list-items don't get the custom item-height applied. Setting item-height="S" or "L" is working, but not with custom height like "60". 

**Suppose**:
`<ix-event-list item-height="60">` with dynamic content

**Expected**:
all ix-event-list-item inside have the same height (60px in rem)

**Actual**:
only initial rendered ix-event-list-item have the height, newly added ones have the default height

### What type of frontend framework are you seeing the problem on?

Angular

### Which version of iX do you use?

v2.6.1

### Code to produce this issue.

```shell
import { Component } from '@angular/core';

@Component({
  selector: 'app-example',
  template: `
  <ix-event-list item-height="60">
  <ng-container *ngFor="let item of items">
    <ix-event-list-item color="color-primary">Text {{item}}</ix-event-list-item>
  </ng-container>
  </ix-event-list>
  <ix-button (click)=onAdd()>Add</ix-button>
  `,
})
export default class EventList {
  items = [1, 2, 3];

  onAdd() {
    this.items.push(this.items.length + 1);
  }
}
```

saumyaupadhyay95:
Hello @danielleroux - Is this issue fixed or still in the works? I'm facing the same issue in a React app. Please see details below:

### **What happened?**
When dynamically changing the content of the `IxEventList` (e.g. updating the data state in React), newly added IxEventListItems don't get the custom `itemHeight` applied. Setting `itemHeight`="S" or "L" is working, but not with custom height like "60".

**Suppose:**
`<IxEventList itemHeight={60} animated={false}>` with dynamic content

**Expected:**
All `<IxEventListItem />` elements that are children of `IxEventList` should have the same height (60px in rem)

**Actual:**
only initial rendered `IxEventListItem` have the height, newly added ones have the default height

### What type of frontend framework are you seeing the problem on?
React

### Which version of iX do you use?
v3.0.0

### Code to produce this issue.

```
<IxEventList itemHeight={60} animated={false}>
    {items.map((item, index) => (
      <IxEventListItem key={index}>{item.title}</IxEventListItem>
     ))}
[...]

PR #1768, opened 2025-03-12
fix(core/event-list): event list custom height applied for dynamic content

RamVinayMandal:

- Custom height in pixels does not apply to dynamically added event list item.

GitHub Issue Number: #1684 

- The custom item height (pixels) is applied to the dynamically added event list item.

## 🏁 Checklist


packages/core/src/components/event-list/event-list.tsx
```diff
@@ -60,8 +61,8 @@ export class EventList {
       this.triggerFadeIn();
     }
 
-    if (!Number.isNaN(Number(this.itemHeight))) {
-      const height = convertToRemString(this.itemHeight as number);
+    if (isNumber(this.itemHeight)) {
```
nuke-ellington, line 64: Why can we not rely on the Number.isNaN. Is there an issue in the current implementation?

packages/core/src/components/event-list/event-list.tsx
```diff
@@ -100,6 +101,10 @@ export class EventList {
   }
 
   private setCustomHeight(item: HTMLElement, height: string) {
+    if (item.nodeType !== Node.ELEMENT_NODE) {
```
nuke-ellington, line 104: What is the purpose of this conditional?

packages/core/src/components/event-list/test/event-list.ct.ts
nuke-ellington, line 1: Test does not test the custom height functionality.
We need to test that all items (including the new one) do have the same custom height.

nuke-ellington (changes requested):
In general we don't create test app example files for all use cases.
We have unit tests for that.

packages/core/src/components/event-list/test/event-list.ct.ts
RamVinayMandal: Thanks for the feedback! I've updated the test to explicitly verify that all event list items, including newly added ones, have the same custom height. The test now waits for styles to be applied and checks the computed height of each item. Let me know if any further adjustments are needed!

packages/core/src/components/event-list/event-list.tsx
RamVinayMandal: The current condition was fine, but there was an incorrect check elsewhere in the same file. To ensure consistency and correctness, I introduced the isNumber utility function. This avoids duplicating logic and makes the checks more reliable and maintainable.

packages/core/src/components/event-list/event-list.tsx
RamVinayMandal: The condition `if (item.nodeType !== Node.ELEMENT_NODE)` ensures that we only apply the custom height to valid HTML elements. This was needed because, in Angular, when using <ng-container>, non-element nodes (like comment nodes from structural directives) can be present in the DOM. The previous approach worked fine in React but caused issues in Angular apps. This condition makes the implementation compatible across Web Components, Angular, React, and Vue by ensuring we only process actual elements.

RamVinayMandal:
> In general we don't create test app example files for all use cases. We have unit tests for that.

I added these preview examples to analyze the bug behavior across Angular, React, Vue, and HTML during development. This use case (event-list-custom-item-height-dynamic-list) is different from the existing one (event-list-custom-item-height). Should I remove the new example, or do you think it adds value for debugging and manual verification?

packages/html-test-app/src/preview-examples/event-list-custom-item-height-in-number.html
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Event List Custom Item Height Dynamic List</title>
```
nuke-ellington, line 6:
```suggestion
    <title>Event List Custom Item Height List</title>
```

packages/html-test-app/src/preview-examples/event-list-custom-item-height-in-number.html
```diff
@@ -0,0 +1,32 @@
...
+    <script type="module">
+      document.addEventListener('DOMContentLoaded', () => {
```
nuke-ellington, line 8: IMO this is way to complicated. Just add the list items directly to the body tag below.

packages/core/src/components/event-list/event-list.tsx
```diff
@@ -79,26 +79,45 @@ export class EventList {
...
+
+    mutationRecords.forEach((mutation) => {
+      if (mutation.type !== 'childList') {
+        return;
+      }
+
+      mutation.addedNodes.forEach((node) => {
+        if (node.nodeType !== Node.ELEMENT_NODE) {
+          return;
+        }
```
nuke-ellington, line 104: Should not be required since we check for tagName directly after.

packages/core/src/components/event-list/event-list.tsx
```diff
@@ -79,26 +79,45 @@ export class EventList {
...
+        const element = node as HTMLElement;
+
+        if (element.tagName === 'IX-EVENT-LIST-ITEM') {
+          eventListItems.push(element);
+        }
+
+        eventListItems.push(
+          ...Array.from(element.querySelectorAll('ix-event-list-item'))
+        );
+      });
```
nuke-ellington, line 115: What is this supposed to do? Nested list items should get their own mutation entry.

packages/core/src/components/event-list/test/event-list.ct.ts
```diff
@@ -68,3 +68,59 @@ test('check if items still clickable', async ({ mount, page }) => {
...
+      <ix-event-list-item color="color-primary">Text 1</ix-event-list-item>
+      <ix-event-list-item color="color-primary">Text 2</ix-event-list-item>
+      <ix-event-list-item color="color-primary">Text 3</ix-event-list-item>
+      <ix-event-list-item color="color-primary">Text 4</ix-event-list-item>
+    </ix-event-list>
+  `);
+
+  const eventListItems = page.locator('ix-event-list-item');
+
+  await expect(eventListItems).toHaveCount(4);
```
nuke-ellington, line 90:
```suggestion
```

packages/core/src/components/event-list/test/event-list.ct.ts
```diff
@@ -68,3 +68,59 @@ test('check if items still clickable', async ({ mount, page }) => {
...
+
+  await page.evaluate(() => {
+    const eventListItem = document.createElement('ix-event-list-item');
+    eventListItem.textContent = 'Newly added item';
+    eventListItem.setAttribute('color', 'color-primary');
+
+    const eventList = document.querySelector('ix-event-list');
+    if (eventList) {
+      eventList.appendChild(eventListItem);
+      const div = document.createElement('div');
```
nuke-ellington, line 100: Why do we add this div here?

packages/core/src/components/event-list/test/event-list.ct.ts
```diff
@@ -68,3 +68,59 @@ test('check if items still clickable', async ({ mount, page }) => {
...
+      div.appendChild(eventListItem);
+      eventList.appendChild(div);
+    }
+  });
+
+  await expect(eventListItems).toHaveCount(5);
+
+  await page.waitForFunction(() => {
+    const lastItem = document.querySelectorAll('ix-event-list-item');
```
nuke-ellington, line 109: This will also succeed with the initial four items, right?!

nuke-ellington (changes requested):

nuke-ellington (changes requested):

packages/core/src/components/event-list/test/event-list.ct.ts
```diff
@@ -68,3 +68,42 @@ test('check if items still clickable', async ({ mount, page }) => {
...
+    const eventList = document.querySelector('ix-event-list');
+    if (eventList) {
+      eventList.appendChild(eventListItem);
+    }
+  });
+
+  const eventListItems = page.locator('ix-event-list-item');
+  await expect(eventListItems).toHaveCount(5);
+
+  await page.waitForTimeout(500);
```
nuke-ellington, line 100: Is the timeout really necessary? We typically try to not use them. Use e.g. `waitForSelector` instead.

nuke-ellington (approved):

PR #1772, opened 2025-03-14
fix(core/date-picker): Add new prop to interface

lzeiml:
As discussed small PR to add the new prop to the Datepicker interface. Also fixed some typos.


danielleroux (dismissed):

danielleroux:
Changeset not needed here its already covered by the minor changeset. I have delete the changeset

danielleroux (approved):

PR #1794, opened 2025-03-24
fix(core/select): clear spacing for inputs in ix-select editable

GayatriK2002:

- Typing a space before a word affects how the value is stored and displayed (e.g., entering "  Item 1" results in "Item 1" being shown).
- In select-editable, entering a space triggers the add icon.

GitHub Issue Number: #

- Leading and trailing spaces are ignored when saving and displaying the value, ensuring consistent selection.
- The add icon only appears when a char (but not space) is entered.

## 🏁 Checklist


packages/core/src/components/select/test/select.ct.ts
```diff
@@ -671,3 +671,54 @@ test('should be 100% when dropdown-max-width is greater than the viewport width'
...
+    <ix-select-item label="Item 7" value="7"></ix-select-item>
+  </ix-select>`);
+
+  const select = page.locator('ix-select');
+  const input = select.locator('Input');
+
+  await input.fill('   Item 7   ');
+  await input.press('Enter');
+
+  await expect(input).toHaveValue('Item 7');
```
raghu-siemens, line 723: can you write one test case more to verify the space in the middle like 'Item           7".

GayatriK2002: Added test case for middle space case

packages/core/src/components/select/test/select.ct.ts
```diff
@@ -671,3 +671,74 @@ test('should be 100% when dropdown-max-width is greater than the viewport width'
...
+    <ix-select-item label="Item 1" value="1"></ix-select-item>
+    <ix-select-item label="Item 2" value="2"></ix-select-item>
+  </ix-select>`);
+
+  const select = page.locator('ix-select');
+  const input = select.locator('input');
+
+  await input.fill('    ');
+
+  const addIcon = page.locator('[isAddItemVisible]');
```
nuke-ellington, line 689: Where is this attribute coming from? `isAddItemVisible`

packages/core/src/components/select/test/select.ct.ts
```diff
@@ -671,3 +671,74 @@ test('should be 100% when dropdown-max-width is greater than the viewport width'
...
+    <ix-select-item label="Item 1" value="1"></ix-select-item>
+    <ix-select-item label="Item 2" value="2"></ix-select-item>
+  </ix-select>`);
+
+  const select = page.locator('ix-select');
+  const input = select.locator('input');
+
+  await input.fill('  Item 1');
+
+  const addIcon = page.locator('[isAddItemVisible]');
```
nuke-ellington, line 707: Same as above

nuke-ellington (changes requested):

packages/core/src/components/select/test/select.ct.ts
GayatriK2002: Removed the attribute and corrected the test case

packages/core/src/components/select/test/select.ct.ts
GayatriK2002: Removed the attribute and corrected the test case

packages/core/src/components/dropdown-item/dropdown-item.tsx
```diff
@@ -118,7 +118,7 @@ export class DropdownItem implements DropdownItemWrapper {
             <ix-icon class="dropdown-item-icon" name={this.icon}></ix-icon>
           ) : null}
           <div class="dropdown-item-text">
-            {this.label}
+            <pre>{this.label}</pre>
```
matthiashader, line 121: I think we shouldn't use a plain pre here, as this will transform the text to a monospace font
image

matthiashader (dismissed):

packages/core/src/components/dropdown-item/dropdown-item.tsx
GayatriK2002: Removed the pre tag and fixed with css

.changeset/healthy-kangaroos-cover.md
nuke-ellington, line 1: Why did we delete the changeset?

GayatriK2002: Pushed that file

packages/core/src/components/select/test/select.ct.ts
```diff
@@ -699,3 +699,77 @@ test('dropdown can be opened after clearing select in editable mode', async ({
...
+  const select = page.locator('ix-select');
+  const input = select.locator('Input');
+
+  await input.fill('   Item 7   ');
+  await input.press('Enter');
+
+  await expect(input).toHaveValue('Item 7');
+});
+
+test('should preserve spaces within input and show add icon', async ({
```
nuke-ellington, line 754: Why is this test so complicated? Can it not be like the one above?

GayatriK2002: Simplified test case as above one.

nuke-ellington (approved):

Issue #1789, opened 2025-03-20
Null value binding form control like ix-input or ix-textarea does not appear as getting error: "Cannot read properties of null (reading 'length')


Jayesh26:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

Hi Team, 

 

We are getting "Cannot read properties of null (reading 'length')" issue on form control fields like "ix-textarea" or "ix-input" having two way binding in angular with null value.

 

In this case the control does not appear on form/screen, attached the snippet.

 

Can we handle null check at ix framework side so that at least form control should be seen on the screen.

 

we are using below version:

 

"@siemens/ix": "2.7.0",

"@siemens/ix-angular": "2.7.0",

"@siemens/ix-icons": "2.2.0",

![Image](https://github.com/user-attachments/assets/5939ffd1-8fe5-451b-8693-0549e3a81b49)

![Image](https://github.com/user-attachments/assets/04eb8396-5988-4117-bdbb-1a52c121b99e)

![Image](https://github.com/user-attachments/assets/3a6ea20b-86a8-432c-a97a-cccb3616a643)

### What type of frontend framework are you seeing the problem on?

Angular

### Which version of iX do you use?

2.7.0

### Code to produce this issue.

```shell
<ix-textarea
          id="text-area-for-{{ fieldName }}"       
          [(ngModel)]=null?
</ix-textarea>
```

danielleroux:
Your issue will be analyzed and is part of our internal workflow.
To get informed about our workflow please checkout the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md#issue-workflow)

JIRA: IX-2526

danielleroux:
Fixed in v3.0.0

PR #1828, opened 2025-04-08
fix(core/textarea): Added optional chaining to prevent runtime errors when null value passed

RamVinayMandal:

When value = null is assigned programmatically to `<ix-textarea>`, the component fails to handle it gracefully. This causes:

- The internal textarea to disappear (nothing rendered).

- Console errors when optional chaining is not used.

GitHub Issue Number: #1789 

- The component now normalizes null values to empty string ('') internally.

- The textarea remains visible even when value = null.

## 🏁 Checklist


packages/core/src/components/input/tests/form-ready.ct.ts
```diff
@@ -86,3 +86,20 @@ test(`form-ready - ix-textarea with initial value`, async ({ mount, page }) => {
...
+    `<form><ix-textarea name="my-field-name" max-length="100"></ix-textarea></form>`
+  );
+
+  await page.evaluate(() => {
+    const el = document.querySelector('ix-textarea');
+    // @ts-ignore
+    el.value = null;
+  });
+
+  await expect(page.locator('textarea')).toBeAttached();
```
danielleroux, line 104: I would suggest to verify your actual change e.g you introduced the `?` inside the component code.
Your test covers only if the textarea is rendered, but I thing you should something like: 

```
expect(...).toBeEqual("0/100")
```

RamVinayMandal: Thanks for the feedback! Implemented the suggested change.

packages/core/src/components/input/tests/form-ready.ct.ts
```diff
@@ -86,3 +86,18 @@ test(`form-ready - ix-textarea with initial value`, async ({ mount, page }) => {
   const formData = await getFormValue(formElement, 'my-field-name', page);
   expect(formData).toBe('initial value');
 });
+
+test('form-ready - textarea correctly renders character counter with null value', async ({ mount, page }) => {
```
nuke-ellington, line 90: Linter fails for this line.

nuke-ellington (changes requested):

packages/core/src/components/input/tests/form-ready.ct.ts
RamVinayMandal: Resolved lint error.

nuke-ellington (approved):

PR #1862, opened 2025-05-04
docs(README): fix link

icyerasor:
The link isn't working / 404.

- the link works

## 🏁 Checklist
only change in readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Adopt major frameworks for seamless integration into your project.
 
 ## 🚀 Getting Started
 
-You can find a getting started guide [here](https://ix.siemens.io/docs/home/overview)
+You can find a getting started guide [here](https://ix.siemens.io/docs/getting-started/developers)
 
 <!-- CONTRIBUTING -->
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -45,7 +45,7 @@ Adopt major frameworks for seamless integration into your project.
 
 ## 🚀 Getting Started
 
-You can find a getting started guide [here](https://ix.siemens.io/docs/installation/)
+You can find a getting started guide [here](https://ix.siemens.io/docs/home/overview)
```
nuke-ellington, line 48: Good catch. I think this is the correct link though :)
```suggestion
You can find a getting started guide [here](https://ix.siemens.io/docs/getting-started/developers)
```

nuke-ellington (changes requested):

nuke-ellington (approved):

README.md
icyerasor: @nuke-ellington 
Not quite.
The developer subpage might be the right one, coming from github - I'm not able to decide this.
The correct link then would be `https://ix.siemens.io/docs/home/getting-started/developers` though. (you omitted the `/home` and thus it seems to be broken again - for me at least).

nuke-ellington: You are right I messed up the link but fixed it on main directly. Should be working now. Thx again.

PR #1867, opened 2025-05-06
Ix 2304 formslot

1307-Dev:

Currently the helper text in form elements only allows text (default, invalid, warning, info, valid). This text does not allow any formatting like line-breaks, bullet points, numbered lists, links, etc. As an example, for password entries it is not possible to use the helper text to provide formatted feedback which criteria are already fulfilled and which not.
GitHub Issue Number: There is no GitHub issue. Jira Issue- https://agileworld.siemens.cloud/jira/browse/IX-2304

- Adaptation of slots to accommodate more than just strings. The same to be displayed based on the state of the form fields.
- Functionality to include customized content in these slots.
- New slots to work in tune with older features like showTextInTooltip, etc.

## 🏁 Checklist

## 👨‍💻 Help & support

As discussed I have created a story in input/slots. I have created a password validator example to be shown to the UX team. After further information, we can add the suggested CSS changes.


packages/core/src/components/input/input.tsx
nuke-ellington, line 1: I think we can get rid of all of the logic expect the actual slots and have the field wrapper handle it.
All @form-ready components need the slots so we want to prevent unnecessary code duplication.

packages/core/src/components/field-wrapper/field-wrapper.tsx
```diff
@@ -107,6 +112,31 @@ export class FieldWrapper implements FieldWrapperInterface {
...
+        statusSlot = <slot name="invalid"></slot>;
+      } else if (this.isWarning && this.hasWarningContent) {
+        statusSlot = <slot name="warning"></slot>;
+      } else if (this.isInfo && this.hasInfoContent) {
+        statusSlot = <slot name="info"></slot>;
+      } else if (this.hasHelperContent) {
+        statusSlot = <slot name="helper"></slot>;
+      }
+      return statusSlot;
+    };
```
nuke-ellington, line 129:
```suggestion

    const renderStatusSlot = () => {
      let statusSlot = 'helper';

      if (this.isValid) {
        statusSlot = 'valid';
      } else if (this.isInvalid) {
        statusSlot = 'invalid';
      } else if (this.isWarning) {
        statusSlot = 'warning';
      } else if (this.isInfo) {
        statusSlot = 'info';
      }

      return <slot name={statusSlot}></slot>;
    };

```

nuke-ellington (changes requested):
Tests and implementation for other @form-ready components (e.g. ix-number-input) is missing.

packages/core/src/components/field-wrapper/field-wrapper.tsx
```diff
@@ -112,31 +107,53 @@ export class FieldWrapper implements FieldWrapperInterface {
...
+      const elements = slot.assignedElements();
+      return elements.some((element) => {
+        if (element.textContent?.trim() || element.children.length > 0) {
+          return true;
+        }
+        if (element.tagName.includes('-')) {
+          return true;
+        }
+
+        return false;
```
nuke-ellington, line 125: What is the `'-'` check for?
```suggestion
        return (element.textContent?.trim() || element.children.length > 0);
```

packages/core/src/components/field-wrapper/field-wrapper.tsx
nuke-ellington, line 1: Please format file

packages/core/src/components/field-wrapper/field-wrapper.tsx
1307-Dev: The '-' check is for when we use empty custom elements. It seems after further testing it was already handled without it, hence I am removed the particular snippet.

danielleroux:
Close due inactivity

PR #1922, opened 2025-05-27
fix(core/chip): Made icon color same as outline color according to Figma

GayatriK2002:

- In chip primary outline variant, chip icon color is having chip text color which is color-contrast-text.
- Custom variant chip example is having different colors from Figma design.

- In chip primary outline variant, chip icon is having same color as chip outline color according to Figma design.
- Custom variant chip example is having same colors as Figma design.

## 🏁 Checklist

## 👨‍💻 Help & support


nuke-ellington (changes requested):
Adapt VRT snapshots

packages/angular-standalone-test-app/src/preview-examples/chip.ts
nuke-ellington, line 1: The examples should be closer to Figma as per the ticket.

packages/core/src/components/chip/chip.tsx
```diff
@@ -134,6 +134,7 @@ export class Chip {
     if (this.variant === 'custom') {
       customStyle = {
         color: this.chipColor,
+        '--border-color': this.background,
```
nuke-ellington: Border color?

packages/core/src/tests/chip/basic/index.html
nuke-ellington, line 1: Should be similar to the examples (compare comment above)

packages/core/src/components/chip/test/chip.ct.ts
nuke-ellington, line 1: Not sure we need this. We have the visual tests already.

packages/core/src/components/chip/chip.tsx
GayatriK2002: The --border-color CSS variable is created to access the custom chip's background color in chip-mixin.scss. When a chip is in outline variant, we need to use the same color for the border as the background color defined in the custom variant. This variable enables sharing of the color value between background and border properties while maintaining consistent styling

packages/angular-standalone-test-app/src/preview-examples/chip.ts
GayatriK2002: I've updated the custom variant colors in the preview examples according to Figma and confirmed these changes with the UX BLR team.
Would you like me to:
Modify the examples to match the single-column layout shown in the ticket?

Please clarify which aspect of the Figma design should be prioritized for these preview examples.

packages/core/src/tests/chip/basic/index.html
GayatriK2002: I added an icon to one of the existing outline variant chips in the visual test specifically to validate the fix for icon-outline color matching.
Could you please clarify:

Should the visual test cases exactly mirror the preview examples?

packages/core/src/components/chip/test/chip.ct.ts
GayatriK2002: Yes, removed it

packages/angular-standalone-test-app/src/preview-examples/chip.ts
nuke-ellington: As discussed in our sync please consult UX to help you with the layout. It should give a useful overview of the available options without overwhelming the user. The Figma design can be an inspiration, but coping it 1:1 will probably be too much.

packages/core/src/tests/chip/basic/index.html
nuke-ellington: As discussed in our sync we need to test all variations. You can mirror the Figma layout if you want.

packages/core/src/components/chip/chip.tsx
nuke-ellington: But it is used to color the icon not the border, right?

GayatriK2002: Yes, But as the icon color should match to border color, I named that variable as --border-color. Shall I change it to 
"--icon-color" ?

packages/angular-standalone-test-app/src/preview-examples/chip.ts
GayatriK2002: Yes, I have updated all the preview examples after discussing with UX team. Now example has all available options.

packages/core/src/tests/chip/basic/index.html
GayatriK2002: Yes, Updated Visual test that checks all variations.

packages/core/src/components/chip/chip.tsx
nuke-ellington: We should not set the color via custom property in the chip mixin, since that is also used elsewhere. This would risk bleeding of 3rd party custom props into it. Please set the color on the icon directly. Try something like this:
`color={this.variant === 'custom' ? this.background : undefined}`

GayatriK2002: Yes, removed '--border-color': this.background'

nuke-ellington (approved):

PR #1941, opened 2025-06-09
fix(modal): animation is always identical regardless of position (#2399)

Raghavendra5577:

Currently, when centered: true is set, the modal slides in from the bottom to the center of the screen. The animation direction does not always match the desired behavior of sliding in from above, especially when the modal is centered.

When centered: true is set, the modal now always slides in from above (top of the viewport) to its centered position, instead of sliding in from the bottom. This ensures a consistent and expected animation direction for centered modals.

-
-

## 🏁 Checklist


packages/core/src/components/modal/modal.tsx
nuke-ellington, line 1: Revert this change please

packages/core/src/components/modal/test/modal.ct.ts
nuke-ellington, line 1: Why are we testing this, when we already have visual regression tests for it?

packages/core/src/components/modal/modal.scss
```diff
@@ -120,13 +130,12 @@
...
-  dialog {
-    margin: 0;
-    left: 50%;
-    top: 50%;
-  }
+:host(.align-center) dialog,
+dialog.align-center {
+  animation: ix-modal-slide-in-top 0.3s cubic-bezier(0.4, 0, 0.2, 1);
+  left: 50%;
+  top: 6rem;
```
nuke-ellington: With this change the dialog will not be centered vertically anymore.

nuke-ellington (changes requested):

packages/core/src/components/modal/modal.scss
```diff
@@ -120,13 +130,12 @@
...
 }
-
-:host(.align-center) {
-  dialog {
-    margin: 0;
-    left: 50%;
-    top: 50%;
-  }
+:host(.align-center) dialog,
+dialog.align-center {
```
nuke-ellington: Why do we need this second selector?

packages/core/src/components/modal/test/modal.ct.ts
Raghavendra5577: Hi Lukas,
I have removed the test cases for this

packages/core/src/components/modal/modal.scss
```diff
@@ -120,13 +130,12 @@
...
+  animation: ix-modal-slide-in-top 0.3s cubic-bezier(0.4, 0, 0.2, 1);
```
nuke-ellington:
```suggestion
  animation: ix-modal-slide-in-top var(--theme-slow-time) cubic-bezier(0.4, 0, 0.2, 1);
```

RamVinayMandal: Handled the animation with anime method from animejs.

RamVinayMandal: > 

Handled the animation with anime method from animejs.

nuke-ellington (approved):

nuke-ellington (changes requested):
Looks like visual tests where affected by the change. Please check.

PR #1947, opened 2025-06-10
fix-1613/ix-toggle-label-do-not-truncate-break-on-overflow

Nadim901:

 - In toggle button  label doesn't truncate / break on overflow

GitHub Issue Number: 683
Jira Issue number: [IX-1613]

- In toggle button  label is now truncate or break on overflow 
- The ix-toggle label text is now wrapping within the container, text and toggle are top-aligned, text breaks down

## 🏁 Checklist


packages/core/src/components/toggle/toggle.scss
```diff
@@ -27,12 +28,15 @@
...
     align-items: flex-start;
     width: 100%;
-    height: 100%;
+    max-height: 100%;
   }
 
   .label {
-    display: inline-block;
+    word-break: break-word;
+    display: -webkit-box;
```
nuke-ellington: Why are we using -webkit-box here??

Nadim901: - We're using -webkit-box display property along with -webkit-line-clamp.
- We are calculating -webkit-line-clamp  dynamically in toggle.txs file to ensure clean text truncation with proper ellipsis.
- This way we truncate text at a specific number of lines with an ellipsis as height is dynamic (coming from parent element)

nuke-ellington: Feedback form UX: feedback via elipsis is not required. Breaking lines as stated in the ticket will be enough.

Nadim901: ok, Done the changes

nuke-ellington (changes requested):
Please remove all of the TS logic - we only want to break the label if there is not enough space

Nadim901:
> Please remove all of the TS logic - we only want to break the label if there is not enough space

Changes done.

nuke-ellington (approved):

PR #1950, opened 2025-06-14
Fix 2621/modal with close backdrop click closes mouse held down inside modal released outside

Nadim901:

- Modal with close-on-backdrop-click closes when mouse is held down inside modal and released outside

GitHub Issue Number: 1663
Jira Issue number: [IX-2621]

- Modal stays open when mouse is held down inside modal and released outside, only closes when "full click" happens outside  the modal

## 🏁 Checklist


packages/core/src/components/modal/modal.tsx
```diff
@@ -137,17 +137,18 @@ export class Modal {
...
     }
+    if (!this.isClickInsideDialog && this.closeOnBackdropClick) {
+      this.dismissModal();
+    }
+  }
 
+  private onMouseDown(event: MouseEvent) {
     const rect = this.dialog!.getBoundingClientRect();
-    const isClickOutside =
+    this.isClickInsideDialog =
```
nuke-ellington, line 147: Do we even need to check the dimensions, since we konw the event comes from dialog element?

Nadim901:
> In storybook and html-test-app modal still closes when releasing over modal after mouse down on backdrop

@nuke-ellington 

 Regarding the this modal backdrop click issue we have two possible approaches for handling mouse events:

Approach 1:
- DOWN inside + UP inside = Stay open
- DOWN inside + UP outside = Stay open
- DOWN outside + UP inside = Close modal
- DOWN outside + UP outside = Close modal

Approach 2 :
- DOWN inside + UP inside = Stay open
- DOWN inside + UP outside = Stay open
- DOWN outside + UP inside = Stay open
- DOWN outside + UP outside = Close modal

The key difference is: should we close ( or stay open) the modal when mousedown starts outside but mouseup happens inside?
**Our current fix is according to Approach 1.**
Just wanted to confirm which Approach we should follow?

nuke-ellington:
> > In storybook and html-test-app modal still closes when releasing over modal after mouse down on backdrop
> 
> @nuke-ellington
> 
> Regarding the this modal backdrop click issue we have two possible approaches for handling mouse events:
> 
> Approach 1:
> 
> * DOWN inside + UP inside = Stay open
> * DOWN inside + UP outside = Stay open
> * DOWN outside + UP inside = Close modal
> * DOWN outside + UP outside = Close modal
> 
> Approach 2 :
> 
> * DOWN inside + UP inside = Stay open
> * DOWN inside + UP outside = Stay open
> * DOWN outside + UP inside = Stay open
> * DOWN outside + UP outside = Close modal
> 
> The key difference is: should we close ( or stay open) the modal when mousedown starts outside but mouseup happens inside? **Our current fix is according to Approach 1.** Just wanted to confirm which Approach we should follow?

UX decision: **Approach 2** should be implemented 👍

packages/core/src/components/modal/modal.tsx
```diff
@@ -34,7 +34,8 @@ export type IxModalSize = IxModalFixedSize | IxModalDynamicSize;
 })
 export class Modal {
   private ariaAttributes: A11yAttributes = {};
-  private isClickInsideDialog = false;
+  private isMouseUpInsideDialog = false;
```
nuke-ellington: This should not be a class member but rather local to the fn it is used in.

nuke-ellington (changes requested):

packages/core/src/components/modal/modal.tsx
Nadim901: Both isMouseDownInsideDialog and isMouseUpInsideDialog are part of the same modal interaction logic, so keeping them as class members ensures consistency and clearer code intent. However, if performance is a priority, I can make isMouseUpInsideDialog local, though it would sacrifice code consistency for minimal gain.

nuke-ellington: Please make it local 🙏

nuke-ellington (approved):

Issue #1486, opened 2024-09-18
Customizable ix-group context menu


RawalA:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [X] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### Suggestion / feature request

Your suggestion here
![image 487569291](https://github.com/user-attachments/assets/2a7d4787-d86b-4140-8997-7d65189d13da)
 
ix-group context menu should be replaced by any other input control such as checkbox. As of now there is no provision to replace ix-group context-menu to any other input control and also even if its not used it is taking extra space.

RawalA:
Hi Team , Is there any update on above issue ?

nuke-ellington:
Hello @RawalA, we will start work on this in the next sprint. Thank you for your patience!

PR #1985, opened 2025-06-30
Modified group to not occupy space for context menu when no slot used

varun-srinivasa:

when group used as per docu, context menu is occupying empty space even though not configured

GitHub Issue Number: #1486 

Context menu doesnot  occupy space when not configured

## 🏁 Checklist

### DIFF
--- a/packages/core/src/components/group/group.tsx
+++ b/packages/core/src/components/group/group.tsx
@@ -91,6 +91,8 @@ export class Group {
 
   @State() showExpandCollapsedIcon = false;
 
+  @State() hasDropdown = false;
+
   private observer: MutationObserver = null!;
 
   @Watch('selected')
@@ -182,11 +184,27 @@ export class Group {
     }
   }
 
+  private checkDropdownSlot() {
+    const slot = this.hostElement.querySelector('slot[name="dropdown"]');
+    if (slot) {
+      const assigned = (slot as HTMLSlotElement).assignedElements({
+        flatten: true,
+      });
+      this.hasDropdown = assigned.length > 0;
+    } else {
+      this.hasDropdown = !!this.hostElement.querySelector('[slot="dropdown"]');
+    }
+  }
+
   componentWillRender() {
     this.groupItems.forEach((item, index) => {
       item.selected = index === this.index;
       item.index = index;
     });
+  }
+
+  componentWillLoad() {
+    this.checkDropdownSlot();
   }
 
   componentDidLoad() {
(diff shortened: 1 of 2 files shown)


packages/core/src/components/group/group.tsx
```diff
@@ -182,13 +184,29 @@ export class Group {
...
+  private checkDropdownSlot() {
+    const slot = this.hostElement.querySelector<HTMLSlotElement>(
+      'slot[name="dropdown"]'
+    );
+    if (slot) {
+      this.hasDropdown = slot.assignedElements({ flatten: true }).length > 0;
+    } else {
+      this.hasDropdown =
+        this.hostElement.querySelector('[slot="dropdown"]') !== null;
+    }
```
nuke-ellington: There is a util fn called `hasSlottedElements` that should be used. Also the fallback should not be necessary here since the named slot is available.

packages/core/src/components/group/test/group.ct.ts
nuke-ellington, line 1: Since this change is evident in the visual tests we don't need component tests for this IMO.

packages/core/src/components/group/group.tsx
varun-srinivasa: This method checks for the presence of elements with [slot="dropdown"] in the light DOM of the ix-group component. This approach is for determining if dropdown content has been provided or dropdown slot was used by the consumer, as we only care about the presence . Using [hasSlottedElements] is not necessary here, since we are not checking a shadow DOM slot but rather looking for slotted elements directly in the component’s light DOM.
For your current use case we anyhow have to check for slot in the light DOM. And yes fallback method was not required in here.

packages/core/src/components/group/test/group.ct.ts
varun-srinivasa: Yes removed the component tests

packages/core/src/components/group/group.tsx
```diff
@@ -217,6 +206,8 @@ export class Group {
     this.observer.observe(this.groupContent, {
       childList: true,
     });
+    //for initial check
+    this.checkDropdownSlot();
```
nuke-ellington, line 210: Do we really need to check on will load **AND** did load?

packages/core/src/components/group/group.tsx
```diff
@@ -199,6 +206,8 @@ export class Group {
     this.observer.observe(this.groupContent, {
       childList: true,
     });
+    //for initial check
```
nuke-ellington:
```suggestion

```

nuke-ellington (changes requested):

nuke-ellington (approved):

Issue #1982, opened 2025-06-26
ix-textarea strings displayed as objects


siemandalorian:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

Created from community: https://community.siemens.com/t/ix-textarea-strings-displayed-as-objects/2804

We recently upgraded ix-angular to v3 and with that replaced all textarea with ix-textarea components. But we faced a problem that even when passing the value as string to ix-textarea  (ex: "{"test": "sample data" }" which is a json stored as string it was always shown inside the textarea as [object Object].

```
        <ix-textarea
          textarea-height="100px"
          textarea-width="100%"
          #credentialsTextArea
          [value]="getFormattedData()"
        ></ix-textarea>
```

We overcome it by replacing all double quotes with single quotes but be need to modify the text for that.

Can you advise how can we display value like: {"test": "sample data" } inside ix-textarea ?

### What type of frontend framework are you seeing the problem on?

Angular

### On which version of the frontend framework are you experiencing the issue?

@angular/core@19.2.3

### Which version of iX do you use?

v3

### Code to produce this issue.

```shell
<ix-textarea
          textarea-height="100px"
          textarea-width="100%"
          [value]="{"attribute1": "example1"}"
        ></ix-textarea>
```

danielleroux:
@siemandalorian This is the default behaviour of javascript object.

```
[value]="aObject"
```
We need to call "toString()" on the object to get an string to render.

If you need to render the object in json format I would recommend to use a pipe e.g. https://angular.dev/api/common/JsonPipe

siemandalorian:
@danielleroux this is what does not work even if I pass the string to ix-textarea. I checked the type before assigning the value and although it is a string still rendered inside textarea as object.

<img width="407" alt="Image" src="https://github.com/user-attachments/assets/75bef860-f67d-424d-84f3-989e47d1d771" />

danielleroux:
@siemandalorian then please create a example (repo or stackblitz) which shows the issue on that base we are analyzing the issue then. Thank you

MasonVX:
@danielleroux Just came here to report the same thing. And it is also the same for normal input fields. It did work before 3.0.

[...]

PR #1990, opened 2025-07-01
fix(core/input): input string validation (#1982)

ridvandmrc:

- String '{}' showed like [object]

GitHub Issue Number: #1982 

- It is working as expect

## 🏁 Checklist

### DIFF
--- a/packages/core/src/components/input/input.tsx
+++ b/packages/core/src/components/input/input.tsx
@@ -217,6 +217,10 @@ export class Input implements IxInputFieldComponent<string> {
     this.value = value;
   }
 
+    getValue(): string {    
+    return  (typeof this.value as unknown) === 'string' ? this.value : JSON.stringify(this.value);
+  }
+
   /** @internal */
   @Method()
   async getAssociatedFormElement(): Promise<HTMLFormElement | null> {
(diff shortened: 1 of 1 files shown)


packages/core/src/components/input/input.tsx
```diff
@@ -217,6 +217,10 @@ export class Input implements IxInputFieldComponent<string> {
     this.value = value;
   }
 
+    getValue(): string {    
+    return  (typeof this.value as unknown) === 'string' ? this.value : JSON.stringify(this.value);
```
danielleroux: 1. You can move the value check inside the function component file input.fc.ts this fc will be used by other components also which avoids duplicated code.
2. Same counts for the textarea.fc
3. Maybe we can add a more general approach e.g

```typescript
function isPrimitive(value: unknown) {
  if(value === null){
      return true;
  }
  
  return !(typeof value == "object" || typeof value == "function")
}

if (!isPrimitive(props.value)) {
    valueForInput = JSON.stringify(props.value);
}
```

Thank you 👍

//Edit

As @nuke-ellington already mentioned some component tests for input and textarea would be nice.

danielleroux:
@ridvandmrc Do you plan to update the PR?

Following points:

- Fix pipeline
- Update merge latest main
- Apply input validation for ALL input related components e.g ix-time-input also

h4de5:
what is the state of this PR? will this be picked up by IX?

there are currently several strings that cannot be written inside any IX input field. e.g. any valid JSON is converted to "[object Object]". strings like: ["text", 123, "true"] are converted to [text,123,true] and so on..

danielleroux:
@ridvandmrc I close this PR due inactivity. Feel free to open a new one.

Issue #1811, opened 2025-03-31
[ix-input] Event that triggers on enter key


alex50105:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### Suggestion / feature request

Currently there is no way (at least what I found) to react on enter key being pressed. On normal html input control I can use the "change" event, which also triggers on the enter key. But the ix-input events (ixBlur, valueChanged) don't trigger on enter and "change" is not supported. So my suggestion is to add support for the standard html change event which should behave the same as on normal html inputs. But another solution that would allow to react on enter would also be ok.

Use case: value submit on enter, e.g. for a search box.
Currently trying to change from an html input to ix-input and I am blocked by this.

danielleroux:
As workaround *if suitable* you can use the `getNativeInputElement` function of the ix-input element to get the inner input. This makes it possible to listen to the standard events.

Depending on which framework you could build a small utility around the component.

alex50105:
@danielleroux thanks, that workaround works (using ViewChild in Angular).
Should forwarding of that event still be added? It would make sense to not need such workarounds

danielleroux:
> 🤖 Hello [@alex50105](https://github.com/alex50105)
> 
> Your issue will be analyzed and is part of our internal workflow. To get informed about our workflow please checkout the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md#issue-workflow)
> 
> JIRA: IX-2551

@alex50105 yes its tracked in our internal backlog, btw you can also use the "keypress" event and check  with `if(event.key === 'Enter')` if the enter key is pressed

alex50105:
> btw you can also use the "keypress" event and check with `if(event.key === 'Enter')` if the enter key is pressed

that is what I do in the ViewChild setter, for lost focus is still rely on normal `ixBlur` event

alex50105:
Bump, this is stil relevant

alex50105:
bump

alex50105:
bump

PR #2007, opened 2025-07-09
Added ixchange event to ix-input, ix-numberinput and ix-textarea

1307-Dev:

There is no event in ix-input which behaves like the native html change event. 

GitHub Issue Number: #1811 

On clicking enter, an ixchange event is fired. Furthermore, when clicking outside the text box wile the value inside the input is changed also causes the event to be sent, similar to the native change input.

## 🏁 Checklist

### DIFF
--- a/packages/core/src/components/tree/tree.tsx
+++ b/packages/core/src/components/tree/tree.tsx
@@ -361,7 +361,8 @@ export class Tree {
       return;
     }
 
-    const id = target.getAttribute('data-tree-node-id');
+    let treeNodeElement = target;
+    const id = treeNodeElement.getAttribute('data-tree-node-id');
     if (!id) {
       return;
     }
@@ -377,7 +378,9 @@ export class Tree {
     }
 
     if (!event.defaultPrevented) {
-      Object.values(this.context).forEach((c) => (c.isSelected = false));
+      for (const context of Object.values(this.context)) {
+        context.isSelected = false;
+      }
       const context = this.getContext(id);
       context.isSelected = true;
       this.setContext(id, context);


lzeiml:
Please make sure that the pipeline is passing. In this case run pnpm lint and fix everything marked as **error**.

packages/core/src/components/input/input.fc.tsx
```diff
@@ -59,6 +60,8 @@ export function TextareaElement(
         width: props.textareaWidth,
       }}
       {...props.ariaAttributes}
+      onChange={() => props.onChange()}
+      onKeyDown={(e) => e.key === 'Enter' && !e.shiftKey && props.onChange()}
```
nuke-ellington: Why filter for shift key here?

1307-Dev: I added it so that linebreaks do not cause ixchange event to trigger.

packages/core/src/components/input/textarea.tsx
```diff
@@ -158,6 +158,11 @@ export class Textarea implements IxInputFieldComponent<string> {
    */
   @Event() ixBlur!: EventEmitter<void>;
 
+  /**
+   * Event emitted when the textarea value is committed (e.g., on blur or enter key)
```
nuke-ellington: Not quite true? Blur alone will not trigger the event, right?

packages/core/src/components/select/select.tsx
```diff
@@ -503,15 +517,29 @@ export class Select implements IxInputFieldComponent<string | string[]> {
...
 
-    if (event.detail) {
       this.inputElement?.focus();
       this.inputElement?.select();
-
       this.removeHiddenFromItems();
       this.isDropdownEmpty = this.isEveryDropdownItemHidden;
     } else {
+      // Emit ixChange if value changed during interaction
+      const currentValue = Array.isArray(this.value)
```
nuke-ellington: Code duplication

packages/core/src/components/select/select.tsx
```diff
@@ -503,15 +517,29 @@ export class Select implements IxInputFieldComponent<string | string[]> {
...
+        ? [...this.value]
+        : this.value;
+
+      if (JSON.stringify(currentValue) !== JSON.stringify(this.previousValue)) {
```
nuke-ellington: Do we really need to stringify the values?

packages/core/src/components/input/textarea.tsx
1307-Dev: Correct, the event will be triggered when there's a change before committing. I will update this info.

packages/core/src/components/select/select.tsx
1307-Dev: I added the stringify's initially as a comparison check. I will add a method which is more comprehensive and use that instead.

packages/core/src/components/date-input/date-input.tsx
```diff
@@ -288,31 +296,36 @@ export class DateInput implements IxInputFieldComponent<string | undefined> {
...
-    }
+    const newValue = handleValueChange(value, this.oldValue, this.ixChange);
 
-    if (!this.format) {
+    if (newValue !== value) {
       return;
     }
+    this.oldValue = newValue;
+    this.value = newValue;
+    this.valueChange.emit(newValue);
```
danielleroux: Little bit confusing naming/code here function to handle emit called handleValueChange. In line 306 you call another emitter but directly which is also called valueChange. I think you need to find some better naming.

packages/core/src/components/date-input/date-input.tsx
```diff
@@ -288,31 +296,36 @@ export class DateInput implements IxInputFieldComponent<string | undefined> {
...
+      this.isInputInvalid = !date.isValid || date < minDate || date > maxDate;
+
+      if (this.isInputInvalid) {
+        this.invalidReason = date.invalidReason || undefined;
+        this.from = undefined;
+      } else {
+        this.updateFormInternalValue(newValue);
+        this.from = newValue;
+        this.closeDropdown();
+      }
```
danielleroux: Why is this refactoring necessary? Don't see any direct need imo.

packages/core/src/components/input/input.util.ts
```diff
@@ -176,7 +177,7 @@ export const addDisposableChangesAndVisibilityObservers = (
   };
 };
 
-function observeElementUntilVisible(
+export function observeElementUntilVisible(
```
danielleroux: Why did you export this function?

packages/core/src/components/input/number-input.tsx
```diff
@@ -307,18 +312,34 @@ export class NumberInput implements IxInputFieldComponent<number> {
...
+                const event = this.ixChange.emit(newValue);
+                if (event.defaultPrevented) {
+                  input.value = this.value.toString();
+                  this.updateFormInternalValue(this.value);
+                } else {
+                  this.value = newValue;
+                  this.updateFormInternalValue(newValue);
+                  this.valueChange.emit(newValue);
+                }
+              }}
```
danielleroux: Spend a additional function as member to avoid massive tsx tree

packages/core/src/components/input/input.util.ts
```diff
@@ -191,3 +192,29 @@ function observeElementUntilVisible(
...
+  emitter: EventEmitter<T>
+): T {
+  if (newValue !== oldValue) {
+    const event = emitter.emit(newValue);
+    if (event.defaultPrevented) {
+      return oldValue;
+    }
+  }
+  return newValue;
+}
```
danielleroux: I’d suggest not baking the “revert to old value if prevented” policy into this util. That decision should live with the caller. 
The caller decides how to handle prevented, instead of this util enforcing “oldValue” automatically.

packages/core/src/components/input/tests/validation.ct.ts
```diff
@@ -39,6 +39,61 @@ test.describe('validation', () => {
...
+    function handleIxChange(e) {
+      window.ixChangeDetail = e.detail;
+    }
+
+    if (ixInput) {
+      ixInput.addEventListener('ixChange', handleIxChange);
+    }
+  `;
+
+      await page.evaluate(setupEventListener);
```
danielleroux: Please checkout other tests. This case of setup the listener is not a good style.

packages/core/src/components/input/textarea.tsx
```diff
@@ -270,6 +275,22 @@ export class Textarea implements IxInputFieldComponent<string> {
...
+
+                if (event.defaultPrevented) {
+                  textarea.value = this.value;
+                  this.updateFormInternalValue(this.value);
+                } else {
+                  this.value = newValue;
+                  this.updateFormInternalValue(newValue);
+                  this.valueChange.emit(newValue);
+                }
+              }}
```
danielleroux: Spend additional function to keep the tsx tree lean.

packages/core/src/components/date-input/date-input.tsx
1307-Dev: I will update the function name.

packages/core/src/components/input/input.util.ts
1307-Dev: Reverting it

packages/core/src/components/input/input.util.ts
1307-Dev: Understood, refactoring the function.

packages/core/src/components/input/tests/validation.ct.ts
1307-Dev: I have modified the test case

packages/core/src/components/input/number-input.tsx
1307-Dev: Removed massive tsx tree, and used a function

packages/core/src/components/input/textarea.tsx
1307-Dev: Removed massive tsx tree, and used a function

packages/core/src/components/input/tests/validation.ct.ts
```diff
@@ -49,16 +49,17 @@ test.describe('validation', () => {
...
-        }
+        };
         const ixInput = document.querySelector('ix-input');
         if (ixInput) {
-          ixInput.addEventListener('ixChange', handleIxChange);
+          ixInput.addEventListener('ixChange', window.handleIxChange);
         }
-      });
+      `;
+      await page.evaluate(handlerScript);
```
danielleroux: Still no javascript string evaluate this is not necessary

danielleroux (changes requested):

packages/core/src/components/input/tests/validation.ct.ts
1307-Dev: There was a nesting function warning in ts, which is why i modified it this way, reverting it.

1307-Dev:
@nuke-ellington @danielleroux As discussed, closing this PR and opening new one. Furthermore removing prevent default behavior as the native change does not support it and fires the event regardless.

PR #2033, opened 2025-07-28
fix(core/category-filter): adjust padding if no icon is shown

1307-Dev:

There is a space appearing when hide-icon property is set to true. 

JIRA issue- [IX-2082](https://agileworld.siemens.cloud/jira/browse/IX-2082)

Said space is removed now an replace with a tiny-space

## 🏁 Checklist


packages/core/src/components/category-filter/category-filter.scss
```diff
@@ -69,7 +69,7 @@
     padding: 1px $large-space 1px 1.75rem !important;
 
     &.no-icon {
-      padding-left: $tiny-space;
+      padding-left: $tiny-space !important;
```
nuke-ellington: Can we find a solution that does not use `!important`? Probably by getting rid of the other one from l69 as well.

1307-Dev: Getting rid of both !important causes this to happen
image
It seems that @include element-input('false') in line 42 overrules all paddings set in different conditions unless !important is used.

packages/core/scss/mixins/_input.scss
```diff
@@ -9,12 +9,16 @@
 @import './fonts';
 @import './validation/form-component';
 
-@mixin element-input($feature-read-only: true) {
+@mixin element-input($feature-read-only: true, $padding: null) {
```
nuke-ellington:
```suggestion
@mixin element-input($feature-read-only: true, $padding: 0.25rem 0.5rem) {
```

packages/core/scss/mixins/_input.scss
```diff
@@ -9,12 +9,16 @@
...
   & {
     @include typography-body;
     min-height: 2rem;
     width: auto;
-    padding: 0.25rem 0.5rem;
+    @if $padding {
+      padding: $padding;
+    } @else {
+      padding: 0.25rem 0.5rem;
+    }
```
nuke-ellington:
```suggestion
    padding: $padding;
```

packages/core/src/components/category-filter/category-filter.scss
```diff
@@ -38,8 +38,49 @@
...
+        $feature-read-only: true,
+        $padding: $padding-with-icon
+      );
+    }
+
+    &.readonly.no-icon {
+      @include element-input(
+        $feature-read-only: true,
+        $padding: $padding-no-icon
+      );
```
nuke-ellington: This seems overly complex. We can probably combine some of these selectors (e.g. `.no.icon` vs. `:not(.no-icon)`).

packages/core/scss/mixins/_input.scss
1307-Dev: Have added the changes

packages/core/scss/mixins/_input.scss
1307-Dev: I have added the changes

packages/core/src/components/category-filter/category-filter.scss
1307-Dev: I have simplified the changes

nuke-ellington (commented):
Why did all the snapshots change? The padding should be the same?

1307-Dev:
> It is still unclear why the screenshots changed for the filter with icon?
I noticed that in ix-category-filter when the state is set to disabled, there is a mismatch in the size of the box and also the text inside while comparing it to the figma document
 
Current implementation
image
image

Figma Design (here it is mentioned that the component height needs to be 32 pixels and the text is also a bit thinner)

image


With my fix (which is in PR rn), the text is using the intended style, as well as the component height is restored to 32 pixels
image
image

I think the hardcoded implementation from earlier (using !important) was causing this error.

nuke-ellington (changes requested):
We should a no-icon version to the basic VRT snaphot.

danielleroux (approved):

PR #2051, opened 2025-08-05
fix(core/slider): Making slider form ready

1307-Dev:

Currently slider is not a form ready component. It doesn't have styles incorporated for various validation states, nor a support for various feedback texts.

New  changes covered (Requirements listed in the JIRA's comment section) -
Change 1:
Tick and thumb both are at the edge.

Change 2:
Added border radius for track and trace both.

Change 3:
Hover and Active state of thumb does not change it's size.

Change 4:
As discussed with Kathrin, all tick size should be 6px now so reduced the size from 8px to 6px.
Used SVG to render the circles properly.

GitHub Issue Number: #
Jira Issue number: [IX-1519](https://agileworld.siemens.cloud/jira/browse/IX-1519)

Made it form ready, added component tests to confirm the same, updated vrts and snapshots

## 🏁 Checklist

### DIFF
--- a/.changeset/tender-games-drive.md
+++ b/.changeset/tender-games-drive.md
@@ -1,0 +1,7 @@
+---
+"@siemens/ix": patch
+---
+
+Make sure centered **ix-modal** will get displayed in desired place on every render.
+
+Fixes #2190
--- a/packages/core/src/components/modal/modal.scss
+++ b/packages/core/src/components/modal/modal.scss
@@ -46,6 +46,7 @@
     overflow: visible;
     max-height: 80vh;
     pointer-events: all;
+    overflow-wrap: break-word;
   }
 
   .modal-size-360 {
--- a/packages/core/src/components/modal/modal.tsx
+++ b/packages/core/src/components/modal/modal.tsx
@@ -166,6 +166,7 @@ export class Modal {
       );
       this.modalVisible = true;
       dialog.showModal();
+      this.slideInModal();
     } catch (e) {
       console.error('HTMLDialogElement not existing');
     }
@@ -230,10 +231,6 @@ export class Modal {
 
       this.dialogClose.emit(reason);
     });
-  }
-
-  componentDidLoad() {
-    this.slideInModal();
   }
 
   componentWillLoad() {
--- a/packages/core/src/components/modal/test/modal.ct.ts
+++ b/packages/core/src/components/modal/test/modal.ct.ts
@@ -337,3 +337,55 @@ regressionTest.describe('message utils', () => {
     });
   });
 });
+
+regressionTest(
+  'centered modal should remain centered after reopening',
+  async ({ mount, page }) => {
+    await mount(``);
+    await setupModalEnvironment(page);
+
+    const positionTolerance = 5;
+    const dialog = page.locator('ix-modal dialog');
+
+    const openCenteredModal = () =>
+      page.evaluate(() => {
+        const elm = document.createElement('ix-modal');
+        elm.innerHTML = `<div>hi</div>`;
+        (globalThis as typeof globalThis & Window).showModal({
+          content: elm,
+          centered: true,
+        });
+      });
+
+    const dismissCurrentModal = () =>
+      page.evaluate(() => {
+        const modal = document.querySelector('ix-modal');
+        if (modal) {
+          (globalThis as typeof globalThis & Window).dismissModal(modal);
+        }
+      });
+
+    const getModalCenterX = async () => {
+      const box = await dialog.boundingBox();
+      if (!box) {
+        throw new Error('Modal dialog bounding box not found');
+      }
+      return box.x + box.width / 2;
+    };
+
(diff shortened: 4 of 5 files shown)


1307-Dev:
@nuke-ellington I have reopened the Merge request with a new branch as there were too many conflicts on the older one.

packages/core/src/components/slider/slider.tsx
```diff
@@ -34,16 +42,52 @@ function between(min: number, value: number, max: number) {
...
+  /**
+   * Show text below the field component
+   */
+  @Prop() helperText?: string;
+
+  /**
+   * Label for the field component
+   */
+  @Prop() label?: string;
+  /**
```
nuke-ellington, line 71:
```suggestion

  /**
   * Show text below the field component
   */
  @Prop() helperText?: string;

  /**
   * Label for the field component
   */
  @Prop() label?: string;

  /**
```

packages/core/src/components/slider/slider.tsx
```diff
@@ -89,9 +133,10 @@ export class Slider {
 
   /**
    * Show error state and message
+   *
+   * @deprecated Will be removed in 4.0.0. Use invalid class instead.
    */
   @Prop() error?: boolean | string;
-
```
nuke-ellington, line 94: Keep newline

packages/core/src/components/slider/slider.tsx
```diff
@@ -151,6 +223,13 @@ export class Slider {
     this.rangeTraceReference = between(this.min, this.traceReference, this.max);
     this.rangeMin = Math.min(this.min, this.max);
     this.rangeMax = Math.max(this.min, this.max);
+    if (this.value !== undefined) {
+      this.formInternals.setFormValue(this.value.toString());
+    }
+  }
```
nuke-ellington, line 239:
```suggestion
  }
  
```

packages/core/src/components/slider/slider.tsx
```diff
@@ -151,6 +223,13 @@ export class Slider {
     this.rangeTraceReference = between(this.min, this.traceReference, this.max);
     this.rangeMin = Math.min(this.min, this.max);
     this.rangeMax = Math.max(this.min, this.max);
+    if (this.value !== undefined) {
+      this.formInternals.setFormValue(this.value.toString());
```
nuke-ellington: My concern here is, that we set the form value even if the value did not change. I guess this will not trigger any unwanted event, etc. but maybe we can add a separate listener?

RamVinayMandal:
@nuke-ellington

Changes added:
Slider's padding is removed from both end. Padding added from each side in examples to guide the user according to the Figma.
Aligned tooltip and thumb positions.
Aligned tick position with thumb.

Note: Distorted tooltip text issue addressed in this PR https://github.com/siemens/ix/pull/2156

1307-Dev:
> @1307-Dev You changed a lot of the calculation of the slider, compared to the current version some ticks are now off the center if a reference marker is used.
> 
> Can you give me any insights why did you change it at all? I do not see amy reason of changing it, because only form functions should be integrated inside the slider.
> 
> Before: image
> 
> After: image



The main reason we had to change some of the calculations used for the slider was do to the face that there was an offset on both the sides which wasn't seen on the figma design.
image
image


In the new design's figma, there doesn't appear to be any offset
image
So to remove the offset and ensure that the pseudothumb as well as the thumb would align, these changes were added in the calcs.

Without the calculation changes, this is what the slider would have looked like with the new form fields (label, feedback texts)
image


I will look into the ticks which are off center and issue a fix for the same.

RamVinayMandal:
@danielleroux 
Kathrin shared some updates in the JIRA ticket, so we'll switch the PR to Draft mode for now and work through the points raised.

packages/core/src/components/slider/slider.scss
```diff
@@ -210,7 +208,7 @@
     border-radius: 100px;
     top: 50%;
     transform: translateY(-50%);
-    left: calc(var(--tick-value) * 100% - 4px);
+    left: calc(var(--tick-value) * (100% - 0.5rem));
```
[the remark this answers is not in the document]
RamVinayMandal: Tick size is now 6px.
Centered tick to place correctly behind the Thumb. Thumb and tick both are at the edge.
Used SVG to show the circle shape.

packages/core/src/components/slider/slider.tsx
nuke-ellington, line 1: Since tags missing

1307-Dev: Added since tags

packages/storybook-docs/src/stories/input.stories.ts
```diff
@@ -14,6 +14,7 @@ import {
   makeArgTypes,
 } from './utils/generic-render';
 
+
```
nuke-ellington:
```suggestion
```

Issue #2020, opened 2025-07-17
The component library does not respect the "prefers reduced motion" flag


dyeske61283:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

Siemens iX makes broad use of smaller animations on many components.
There are two issues with that:
- There does not seem to be a central setting enable or disable animations (which is more of a general feature request)
- But more importantly: The "prefers reduced motion" flag does not seem to be respected

### What type of frontend framework are you seeing the problem on?

Angular (NgModule)

### On which version of the frontend framework are you experiencing the issue?

Angular V19.2.14

### Which version of iX do you use?

v2.7.0

### Code to produce this issue.

```shell
1. Create a small PoC with any animated iX component e.g. iX-Drawer
2. Open the Developer tools of your browser and open the animations viewer
3. Select or emulate the option to reduce motion in the developer tools or browser settings
4. Animations are still played
```

danielleroux:
@dyeske61283 Thank for bringing up this point.

I tried to find the "reduce motion" settings in the dev tools but couldnt find it.

There are just these controls here:
<img width="238" height="73" alt="Image" src="https://github.com/user-attachments/assets/960963e8-ac00-4d48-bfad-6ef4947823a6" />

dyeske61283:
@danielleroux Thank you for reacting so fast on this!
This is indeed not that easy to find 😅 
There is two ways to find that:
1. The dev tools actually have a command palette, that you can use by pressing CMD/CTRL + P like in VS Code and then select "Emulate CSS prefers-reduced-motion: reduce"
2. In the top tab bar select the context menu with the three dots, then "More tools", then Rendering. There you will find a lot of the options to emulate certain preferences, like color theme, reduced motion, different color modes etc.

And to actually watch and analyze the animations you can use the "Animations" pane, wich is also selectable from the same "More tools" menu.

If there are any other questions for this issue, just tag me again. :) 

PS:
A good starting resource on this whole topic of animation accessibility, I can recommend is https://motion.dev/docs/react-accessibility

dyeske61283:
@danielleroux  This issue was solved by you through #2075, right? If so, I think we can close this issue before letting it go stale.

PR #2075, opened 2025-08-19
feat(core): respect reduce-motion settings

danielleroux:

GitHub Issue Number: #2020 

- Update animejs to latest major version
- Replace hardcoded animation values with core library definitions
- Set all animation values to 0ms if reduce motion is set

## 🏁 Checklist


packages/core/src/components/utils/animation.ts
```diff
@@ -8,9 +8,31 @@
...
+    return this.prefersReducedMotion() ? 0 : 300;
+  }
+
+  static get slowTime() {
+    return this.prefersReducedMotion() ? 0 : 500;
+  }
+
+  static get xSlowTime() {
+    return this.prefersReducedMotion() ? 0 : 1000;
+  }
```
lzeiml: Is it not possible to get the theme variables here instead of hardcoding the values?

danielleroux: Would require to get the computedStyle sheet and get the value from there.

```
const sh = window.getComputedStyle(document.body);
const timing = sh.getPropertyValue('--theme-default-time');
```

One think which needs to be explain is that `getComputedStyle` will resolve a "snapshot" of the css properties.
Which means if the variable --theme-default-time changes after the initial `getComputedStyle` the getter will return the old values. (no big deal imo, because they should never change 😉)

I will do a refactoring.

lzeiml (approved):

PR #2080, opened 2025-08-20
fix(core/category-filter): fixed automatic focus in filter when filterState is set programmatically

GayatriK2002:

- When the filterState property of the category filter component is set programmatically (e.g., through code), the input element automatically receives focus, even if the focus was previously on another element outside the category filter component.
- This behavior occurs regardless of whether the input element was already focused or not.
- The unexpected focus change can disrupt the user experience and cause confusion.

- The category filter component has been updated to prevent the input element from automatically receiving focus when the filterState is set programmatically.

## 🏁 Checklist

## 👨‍💻 Help & support


packages/core/src/components/category-filter/category-filter.tsx
nuke-ellington, line 1: This will only work if the input is already focused. If the the user clicks the dropdown or navigates it via keyboard this will the focus will not work anymore.
Maybe we can extract the focus logic from `addToken` and add it to the respective event handlers only.

nuke-ellington (changes requested):

packages/core/src/components/category-filter/category-filter.tsx
GayatriK2002: Yes, you are right. I have updated the code now

danielleroux (approved):

PR #2105, opened 2025-09-01
fix(core/select): reset filter when items gets selected in multiple selection mode

lakshmi-priya-b:

In ix-select with mode="multiple", when a user filters options by typing and then selects an item from the filtered list
- The filter remains active.
- Only the items that match the filter criteria are shown.

JIRA Ticket: [IX-2446](https://agileworld.siemens.cloud/jira/browse/IX-2446)

With this change, when using ix-select in multiple selection mode, after a user selects an item from a filtered list:
- The input field will automatically clear the filter text.
- The dropdown list will remain open, but will now display all available options again (not just the previously filtered ones).

## 🏁 Checklist


packages/core/src/components/select/test/select.ct.ts
```diff
@@ -120,6 +120,62 @@ test('multiple selection', async ({ mount, page }) => {
...
+  await mount(`
+    <ix-select mode="multiple">
+      <ix-select-item value="1" label="Item 1">Test</ix-select-item>
+      <ix-select-item value="2" label="Item 2">Test</ix-select-item>
+      <ix-select-item value="3" label="Item 3">Test</ix-select-item>
+      <ix-select-item value="4" label="Item 4">Test</ix-select-item>
+    </ix-select>
+  `);
+  const element = page.locator('ix-select');
+  await element.evaluate((select: HTMLIxSelectElement) => (select.value = []));
```
nuke-ellington: Should be [] by default.

packages/core/src/components/select/test/select.ct.ts
```diff
@@ -120,6 +120,62 @@ test('multiple selection', async ({ mount, page }) => {
...
+  await expect(item3).toBeFocused();
+
+  await page.keyboard.press('ArrowUp');
+  await expect(item2).toBeFocused();
+
+  await item3.click();
+  await expect(element.locator('.chips').getByTitle('Item 3')).toBeVisible();
+  await expect(element.locator('input')).toHaveValue('');
+  await expect(dropdown).toBeVisible();
+  await expect(item3).toBeFocused();
```
nuke-ellington: Not really necessary here
```suggestion
```

packages/core/src/components/select/test/select.ct.ts
lakshmi-priya-b: Yes it is [] by default.I have removed and tested.

packages/core/src/components/select/test/select.ct.ts
lakshmi-priya-b: Removed those successfully.

PR #2111, opened 2025-09-03
feat(core/pane): add slot and aria labels

danielleroux:

Jira: IX-3144

- Implement successor features for drawer replacement
- Documentation changes: https://github.com/siemens/ix-docs/pull/84

## 🏁 Checklist


packages/core/src/components/drawer/drawer.tsx
```diff
@@ -22,6 +22,9 @@ import {
 import { animate } from 'animejs';
 import Animation from '../utils/animation';
 
+/**
+ * @deprecated Will be removed with v5, use ix-pane as successor
```
lzeiml: This is how we usually write it, I think we should keep it consistent:

```suggestion
 * @deprecated Will be removed with 5.0.0, use ix-pane as successor
```

packages/core/src/components/pane/pane.tsx
```diff
@@ -170,7 +188,34 @@ export class Pane {
...
+            const path = event.composedPath?.() || [];
+            if (!path.includes(this.hostElement)) {
+              this.dispatchExpandedChangedEvent();
+            }
+          }
+        );
+        return;
+      }
+    }
+    this.disposableWindowClick?.();
```
lzeiml: What do you think about removing the nested ifs here:

```suggestion
    if (!this.closeOnClickOutside || !this.expanded) {
      this.disposableWindowClick?.();
      return;
    }

    this.disposableWindowClick = addDisposableEventListener(
      window,
      'click',
      (event) => {
        const path = event.composedPath?.() || [];
        if (!path.includes(this.hostElement)) {
          this.dispatchExpandedChangedEvent();
        }
      }
    );
```

packages/core/src/components/pane/pane.tsx
```diff
@@ -649,25 +690,30 @@ export class Pane {
...
                 ></ix-icon>
-              ) : null}
+              )}
               <div class="title-text-overflow">
                 <ix-typography format="h4">{this.heading}</ix-typography>
               </div>
-            </span>
+              {this.expanded && (
+                <div class="slot-header">
+                  <slot name="header"></slot>
```
princequator, line 712: Suggestion: Hi, we(Brightly) are planning to use `ix-pane` in our features but have a slightly different requirement regarding the order of elements in the header, as well as adding some buttons there. I just stumbled upon this PR and I noticed a slot being added next to the title in the header, which addresses part of our needs. However, since we want to rearrange the order of the buttons, title, and close button differently, I was wondering if, in addition to the current update, the entire header section could be wrapped in a named slot with keeping original header content as a fallback. This would make the header much more flexible, allowing it to be customized fully through slots. 

PS: If the change can't be addressed in this PR, I'd be happy to contribute and raise a new one.

danielleroux: Thank for your message. The will not be addressed inside this PR. We will get in contact with you via teams.

Issue #2011, opened 2025-07-14
[Feature Request]: Implement standard hyperlink behavior for navigation components


philSixZero:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### Current behavior

Navigation components in Siemens iX (buttons, breadcrumbs, nav entries, tabs, ...) don't support standard browser link functionality:

- No context menu on right-click
- Can't middle-click/Ctrl+click to open in new tab
- No keyboard navigation support matching browser defaults
- Non-bookmarkable routes

### Expected behavior

All clickable navigation elements should implement full HTML hyperlink functionality while maintaining current styling.

For instance, GitHub's & GitLab's UI components demonstrate the correct implementation - all components having navigation purpose are wrapped in an anchor element:

```html
<!-- GitHub-style implementation of navigation items -->
<a href="/path/to/route" class="navigation-item" data-pjax="#js-repo-pjax-container">
  <span class="icon"><!-- SVG icon --></span>
  <span>Navigation text</span>
</a>
```

### Proposal

Introduce a new property `href` throughout all components which have possible navigation purpose:

```html
<ix-button href="/path/to/route">Click me</ix-button>
<ix-breadcrumb-item href="/path/to/section">Section</ix-breadcrumb-item>
<ix-tab-item href="/dashboard">Tab name</ix-tab-item>
```

When `href` is specified, the component should render using an anchor element.

danielleroux:
@philSixZero You idea makes totally sense but currently I dont have a good idea to implement that without any strange workarounds, but any help is welcome!

I would suggest to keep it close to the wrapping approach like with the button

```html
<a href="#" class="ix-no-style">
    <ix-button>A link</ix-button>
</a>

<style>
 .ix-no-style { all: unset; }
</style>

```

<img width="296" height="122" alt="Image" src="https://github.com/user-attachments/assets/e7f4df23-1bee-4472-a70a-567948521693" />

Breadcrumb:

```html
    <ix-breadcrumb>
      <ix-breadcrumb-item>Not wrapped</ix-breadcrumb-item>

      <a href="https://www.google.com" class="ix-no-style">
        <ix-breadcrumb-item href="#">Wrapped</ix-breadcrumb-item>
      </a>
      <ix-breadcrumb-item>Not wrapped</ix-breadcrumb-item>
[...]

PR #2115, opened 2025-09-03
feat(button): add href, target, and rel properties for link functionality

philSixZero:

GitHub Issue Number: #2011

- The IxButton component now renders as an anchor (`<a>`) tag when an `href` property is provided.
- New properties `target` and `rel` were added to support enhanced link behavior.
- Related components (`IxBreadcrumbItem`, `IxMenuItem`) were updated to support the `href` property.

## 🏁 Checklist


danielleroux:
@philSixZero Thank for the contribution

- Button component has `rel` and `target` attribute, but these attributes are missing for breadcrumb-item and menu-item.

Would it make more sense to align these attributes maybe spend a additional interface

```tsx
interface AnchorInterface {
  href: string | undefined;
  target: string | undefined;
  rel: string | undefined;
}
```

philSixZero:
> @philSixZero Thank for the contribution
> 
>     * Button component has `rel` and `target` attribute, but these attributes are missing for breadcrumb-item and menu-item.
> 
> 
> Would it make more sense to align these attributes maybe spend a additional interface
> 
> ```tsx
> interface AnchorInterface {
>   href: string | undefined;
>   target: string | undefined;
>   rel: string | undefined;
> }
> ```

This was done on purpose, as I assumed that breadcrumbs and menu items would consistently link to the same site. However, you might be right that maintaining consistency could be more important here.

Is there a file/folder for common interfaces?

danielleroux:
> > @philSixZero Thank for the contribution
> > ```
> > * Button component has `rel` and `target` attribute, but these attributes are missing for breadcrumb-item and menu-item.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Would it make more sense to align these attributes maybe spend a additional interface
> > ```tsx
> > interface AnchorInterface {
> >   href: string | undefined;
> >   target: string | undefined;
> >   rel: string | undefined;
> > }
> > ```
> 
> This was done on purpose, as I assumed that breadcrumbs and menu items would consistently link to the same site. However, you might be right that maintaining consistency could be more important here.
> 
> Is there a file/folder for common interfaces?

You can place the interface parallel to the component `/src/components/button/button.interface.ts`

packages/core/src/components/button/base-button.tsx
```diff
@@ -99,53 +100,87 @@ export const BaseButton: FunctionalComponent<BaseButtonProps> = (
...
+        href={props.disabled ? undefined : props.href}
+        target={props.target}
+        role="button"
+        rel={props.rel}
+        onClick={(e: Event) => {
+          if (props.disabled || props.loading) {
+            e.preventDefault();
+            e.stopPropagation();
+            return;
+          }
```
danielleroux: Can you give me any insights why handling onClick here different thent he onClick of the button element?

packages/storybook-docs/package.json
```diff
@@ -4,7 +4,7 @@
   "version": "0.0.0",
   "type": "module",
   "scripts": {
-    "build.watch": "storybook dev -p 6006",
+    "build.watch": "storybook dev -p 6007",
```
danielleroux: We should not change the storybook port here.

packages/core/src/components/button/base-button.tsx
philSixZero: First of all, the button's `onClick` callback is simply firing the components `onClick` handler which is not necessary for anchor buttons I would assume.
These lines are needed to deactivate the default click behavior of the anchor element, since the HTMLAnchorElement doesn't have a `disabled` attribute like the HTMLButtonElement (see [MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLButtonElement/disabled)).

philSixZero: If we want to keep this, maybe we should mention somewhere that setting the `href` attribute won't fire the `onClick` handler anymore.

danielleroux: Got it! Via styling we prevent any interaction by `pointer-events: none`:
https://github.com/siemens/ix/blob/ace22d0fd9a126891ef2f6ad60c35751488b81a1/packages/core/src/components/button/button-mixin.scss#L103-L107

//Update:
I think we can handle here the same way like you did also for the button. If the props.disabled or props.loading no click event is necessary

danielleroux:
@philSixZero Regarding the failed build just check your local stash looks like you forgot to add some generated type files or the component-doc.json

tooling/oss-clearing/create-sbom.ts
```diff
@@ -69,12 +69,11 @@ async function resolveComponentData(
...
-    const { stdout } = spawnSync('pnpm', [
-      'v',
-      '--json',
-      '--long',
-      `${packageName}@${version}`,
-    ]);
+    const { stdout } = spawnSync(
+      'pnpm',
+      ['v', '--json', '--long', `${packageName}@${version}`],
+      { shell: true }
```
philSixZero, line 75: @danielleroux stumbled over an `ENOENT` error when trying to build the oss-clearing package; see [this issue](https://github.com/nodejs/node/issues/52554#issuecomment-2060026269).

danielleroux: Alright thank you for pointing out. I assume that you are working on a windows machine without wsl?

You can execute the build with:

```
pnpm build --filter @siemens/ix
```

This will generate all files which are necessary for the PR build. I will update the contribution guides later on.

packages/core/src/components/button/base-button.tsx
```diff
@@ -179,6 +179,7 @@ export const BaseButton: FunctionalComponent<BaseButtonProps> = (
   return (
     <button
       {...commonAttributes}
+      role="button"
```
danielleroux: Do not apply role attribute on the button tag (https://www.w3.org/TR/2014/REC-html5-20141028/forms.html#the-button-element)

image

Maybe find a other selector for the `setFocus`-Method. Maybe do it that way:

```ts
  setFocus() {
    this.hostElement
      .shadowRoot!.querySelector<HTMLButtonElement | HTMLAnchorElement>(
        this.isButtonElement() /* or check for this.href */ ? 'button' : 'a'
      )
      ?.focus();
  }
```

We are currently working on a more "general" solution for focusing elements.

packages/core/src/components/button/button.tsx
```diff
@@ -174,7 +174,11 @@ export class Button implements IxButtonComponent {
   }
 
   setFocus() {
-    this.hostElement.shadowRoot!.querySelector('button')?.focus();
+    this.hostElement
+      .shadowRoot!.querySelector<HTMLButtonElement | HTMLAnchorElement>(
+        '[role="button"]'
+      )
+      ?.focus();
```
danielleroux: See comment from `base-button.tsx:182`

danielleroux (approved):

danielleroux:
@philSixZero Thanks for the contribution!

tooling/oss-clearing/create-sbom.ts
philSixZero: Right, this only happens on Windows. With this change, the error is gone though

PR #2137, opened 2025-09-15
split button focus behavior

varun-srinivasa:

Jira issue number: [IX-3051](https://agileworld.siemens.cloud/jira/browse/IX-3051)

- Added new focus behavior as per figma design
-

## 🏁 Checklist

### DIFF
--- a/packages/core/src/components/split-button/split-button.tsx
+++ b/packages/core/src/components/split-button/split-button.tsx
@@ -100,6 +100,19 @@ export class SplitButton {
   private get dropdownItems(): HTMLElement[] {
     return Array.from(this.hostElement.querySelectorAll('ix-dropdown-item'));
   }
+
+  private get actionButton() {
+    return this.hostElement.shadowRoot?.querySelector(
+      'ix-button, ix-icon-button:not(.anchor)'
+    ) as HTMLElement | null;
+  }
+
+  private get anchorButton() {
+    return this.hostElement.shadowRoot?.querySelector(
+      'ix-icon-button.anchor'
+    ) as HTMLElement | null;
+  }
+
   private onDropdownShowChanged(event: CustomEvent<boolean>) {
     if (event.detail) {
       this.arrowFocusController = new ArrowFocusController(
@@ -116,7 +129,6 @@ export class SplitButton {
       this.dropdownElement?.removeEventListener('keydown', this.handleKeyDown);
     }
   }
-
   private handleKeyDown = (event: KeyboardEvent) => {
     if (event.key !== 'Tab') {
       return;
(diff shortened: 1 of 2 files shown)


packages/core/src/components/split-button/split-button.tsx
```diff
@@ -98,13 +99,79 @@ export class SplitButton {
...
+        anchorButton.removeAttribute('tabindex');
+      });
+    }
+  };
+
+  private focusDropdownItem(index: number) {
+    if (index < 0) return;
+    const items = this.dropdownItems;
+    items.forEach((item, i) =>
+      item.setAttribute('tabindex', i === index ? '0' : '-1')
```
nuke-ellington, line 167: Why do we need to change the tabIndex?

packages/core/src/components/split-button/split-button.tsx
```diff
@@ -98,13 +99,79 @@ export class SplitButton {
...
+      this.arrowFocusController = undefined;
+      this.dropdownElement?.removeEventListener('keydown', this.handleKeyDown);
+    }
+  }
+
+  private handleKeyDown = (event: KeyboardEvent) => {
+    if (event.key !== 'Tab') {
+      return;
+    }
+    this.dropdownItems.forEach((item) => item.setAttribute('tabindex', '-1'));
```
nuke-ellington, line 136: Why do we need to set the tabIndex?

packages/core/src/components/split-button/split-button.tsx
```diff
@@ -98,13 +99,79 @@ export class SplitButton {
...
+    this.dropdownElement!.show = false;
```
nuke-ellington, line 137: Why is this necessary?

packages/core/src/components/split-button/split-button.tsx
```diff
@@ -98,13 +99,79 @@ export class SplitButton {
...
+    if (!event.shiftKey) {
+      return;
+    }
+    const actionButton = this.hostElement.shadowRoot?.querySelector(
+      'ix-button, ix-icon-button:not(.anchor)'
+    ) as HTMLElement | null;
+
+    const anchorButton = this.hostElement.shadowRoot?.querySelector(
```
nuke-ellington: Getter would be better IMO.

nuke-ellington (changes requested):

packages/core/src/components/split-button/split-button.tsx
varun-srinivasa: When the dropdown is closed by pressing the Tab key, we set the tabindex of all dropdown items to -1. This removes them from the natural tab order of the page. It ensures that a user cannot accidentally Tab into the items of a closed dropdown. Focus should move to the next focusable element outside of the split-button component, which is the standard and expected keyboard navigation behavior.

packages/core/src/components/split-button/split-button.tsx
varun-srinivasa: When Shift+Tab is pressed, the browser's default behavior is to move focus to the previous focusable element, which in this case is the split-button's dropdown trigger (the anchorButton).
However, the desired behavior is to close the dropdown and move focus to the element before the entire split-button component.
To achieve this, we temporarily set tabindex="-1" on the anchorButton, making it non-focusable. This allows the browser's default Shift+Tab action to skip the anchorButton and move focus to the correct preceding element.
The tabindex is immediately restored within a requestAnimationFrame callback so that the anchorButton is focusable again for future interactions.

danielleroux:
PR will be closed regarding general refactoring of keyboard navigation https://github.com/siemens/ix/pull/2268

PR #2166, opened 2025-09-24
refactor(core): change all boolean properties to default to false

nuke-ellington:

- 🚧
-

## 🏁 Checklist


packages/core/scss/components/_input-group.scss
danielleroux, line 1: Is there any specific reason why you removed this input-group styles?

packages/core/src/components/drawer/drawer.tsx
```diff
@@ -39,6 +39,7 @@ export class Drawer {
   /**
    * Fired in case of an outside click during drawer showed state
    */
+  // eslint-disable-next-line @stencil-community/ban-default-true
   @Prop() closeOnClickOutside = true;
```
danielleroux, line 43: Why do we have here the eslint-disable annotation?

packages/core/src/components/map-navigation/map-navigation.tsx
danielleroux, line 1: MapNavigation and BasicNavigation should be removed with 4.0.0. I think we can remove the component or will this be done in a other PR?

packages/core/src/tests/utils/test/page.ts
```diff
@@ -130,8 +130,8 @@ export const regressionTest = testBase.extend<{
     page = await extendPageFixture(page, testInfo);
 
     await page.route('*/**/svg/*.svg', async (route, request) => {
-      if (!process.env.CI) {
-        const [__, svg] = request.url().split('/svg/');
+      if (process.env.CI !== 'true') {
```
danielleroux: This check here is not correct. The variable CI is no necessary "true" could also be "1"
`CI=1 pnpm test`
as it was before was totally correct.

packages/html-test-app/src/preview-examples/validation.html
danielleroux, line 1: Do we need this example anymore?

packages/core/src/components/drawer/drawer.tsx
nuke-ellington: Did not want to refactor deprecated drawer

packages/core/src/components/map-navigation/map-navigation.tsx
nuke-ellington: Out of scope for ticket. Will create separate PR!

PR #2193, opened 2025-10-13
fix(ix-input): hide eye icon for disabled password input

alexkaduk:


packages/core/src/components/field-label/field-label.tsx
```diff
@@ -70,12 +74,9 @@ export class FormFieldLabel implements IxComponent {
     }
   }
 
-  componentWillRender() {
-    this.checkForInternalState();
-  }
-
   componentWillLoad(): void | Promise<void> {
     this.a11yAttributes = a11yHostAttributes(this.hostElement);
+    this.checkForInternalState();
```
nuke-ellington: Can you please explain why we moved this from _willRender_ to _willLoad_ which is a one time callback opposed to an on every render hook. Not saying it is wrong, just curious 👍

alexkaduk: Faced infinite updates loop. After this update and with test cases I did, it works as expected. Some motivation details.
 
When checkForInternalState() was in componentWillRender():
1 componentWillRender() runs before every render
2 checkForInternalState() is called, which checks for invalid state
3 inside checkForInvalidState(), we update this.isInvalid (a mutable prop)
4 Stencil automatically triggers a re-render because a mutable prop changed
5 this triggers another render, which calls componentWillRender() again
6 loop might appear

This is what you saw in the browser console logs - the component was continuously toggling between invalid states (and this was the reasoning to update it; in case it can be checked from smb side as well, would be perfect for me).

componentWillLoad() should works better. it only runs once when the component is first initialized. 

Initial Load:
componentWillLoad() → checkForInternalState() → sets up observers

Runtime Updates:
MutationObserver → checkForInvalidState() → updates this.isInvalid → Stencil auto re-renders

PR #2215, opened 2025-10-17
fix(ix-event-list): adjust examples, add itemColor, add on select handler, add filled variant

alexkaduk:
Results are
image
image

### DIFF
--- a/packages/core/src/components.d.ts
+++ b/packages/core/src/components.d.ts
@@ -1539,7 +1539,7 @@ export namespace Components {
          */
         "disabled": boolean;
         /**
-          * Color of the status indicator. You can find a list of all available colors in our documentation. Example values are `--theme-color-alarm` or `color-alarm`  {@link https://ix.siemens.io/docs/theming/colors/}
+          * Color of the status indicator. You can find a list of all available colors in our documentation. Example values are `--theme-color-alarm` or `color-alarm`  {@link https://ix.siemens.io/docs/styles/colors}
          */
         "itemColor"?: string;
         /**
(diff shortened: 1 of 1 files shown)


packages/core/src/components/event-list-item/event-list-item.tsx
```diff
@@ -37,7 +37,7 @@ export class EventListItem {
    * You can find a list of all available colors in our documentation.
    * Example values are `--theme-color-alarm` or `color-alarm`
    *
-   * {@link https://ix.siemens.io/docs/theming/colors/}
+   * {@link https://ix.siemens.io/docs/styles/colors#status-colors}
```
nuke-ellington: No need to link to status colors since custom colors are also allowed
```suggestion
   * {@link https://ix.siemens.io/docs/styles/colors}
```

packages/html-test-app/src/preview-examples/event-list-filled.html
```diff
@@ -0,0 +1,40 @@
...
+    <script type="module">
+      const items = [
+        { text: 'Text 1', color: 'color-primary' },
+        { text: 'Text 2', color: 'color-primary' },
+        { text: 'Text 3', color: 'color-alarm' },
+        { text: 'Text 4', color: 'color-success' },
+      ];
+
+      const eventList = document.getElementById('event-list');
+      if (!eventList) return;
```
nuke-ellington: This will throw `Illegal return statement` error since it is not inside a function

alexkaduk: fix. thanks

nuke-ellington (approved):

packages/vue-test-app/src/preview-examples/application-breakpoints.vue
```diff
@@ -81,10 +73,22 @@ export default {
...
-        <IxRadio value="lg" label="Large"></IxRadio>
+      <IxRadioGroup :value="breakpoints[0]">
+        <IxRadio
+          value="sm"
+          label="Small"
+          @click="setBreakpoint('sm')"
+        ></IxRadio>
+        <IxRadio
+          value="md"
+          label="Medium"
```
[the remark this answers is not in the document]
alexkaduk: valueChange does not work for Vue. Tried diff approaches, however could not used it. to make the example workable, added listener to IxRadio

nuke-ellington (approved):

PR #2218, opened 2025-10-19
fix(ix-textarea): preserve manual resize dimensions during re-renders

alexkaduk:


packages/core/src/components/input/textarea.tsx
nuke-ellington, line 1: We can remove the inline comments

alexkaduk: done. thanks

PR #2241, opened 2025-11-03
test(vue): add component tests for event handling

alexkaduk:
### 🧪 Testing Added

**Why:** No automated tests existed to validate event handling or prevent regressions.

**What:** Added Vitest testing infrastructure with Playwright browser mode:

**`packages/vue/`** - Unit tests for wrapper library

- Configure Vitest with Playwright browser mode
- Add `plugin.spec.ts` - validates plugin initialization & event naming (camelCase preservation)

**`packages/vue-test-app/`** - Component integration tests

- Configure Vitest with Playwright (chromium)
- Add global test setup (`setup.ts`) to initialize `ixPlugin`
- Add `message.spec.ts` - validates modal/message removal from DOM (real component)
- Add `application-breakpoints.spec.ts` - validates radio group event handling

**Result:** Tests fail with broken `ce()` helper, pass with fix.

---

### 🐛 Radio Component Fix: Make `ix-radio` Directly Clickable & Accessible

**Problem:**

1. **Not Clickable in Tests:** Clicking `<ix-radio>` host element did nothing in unit tests - required accessing shadow DOM directly (works as expected in browser)
2. **Missing Accessible Name:** `label` prop wasn't exposed to accessibility tree, breaking Testing Library queries and screen readers

**Solution:** `packages/core/src/components/radio/radio.tsx`

```diff
  <Host
    aria-checked={a11yBoolean(this.checked)}
    aria-disabled={a11yBoolean(this.disabled)}
+   aria-label={this.label}
    role="radio"
+   onClick={() => !this.disabled && this.setCheckedState(true)}
    onBlur={() => this.ixBlur.emit()}
  >
```

**Changes:**

1. **`aria-label={this.label}`** - Exposes label to accessibility tree

   - Enables `screen.findByRole('radio', { name: 'Small' })`
   - Screen readers can now announce the label
   - Follows [WAI-ARIA best practices](https://www.w3.org/WAI/ARIA/apg/patterns/radio/)

2. **`onClick` handler** - Makes entire host element clickable in tests
   - Enables Testing Library queries to click the radio directly
   - Uses `setCheckedState(true)` (radios should only be checkable)
   - Respects `disabled` state

**Testing:**

```typescript
// ❌ Before: Required shadow DOM access
const button = radio.shadowRoot.querySelector('button');
button.click();

// ✅ After: Direct click like a real user
const radio = await screen.findByRole('radio', { name: 'Small' });
radio.click();
```

---

### 🔄 Vue Component Update

**`packages/vue-test-app/src/preview-examples/application-breakpoints.vue`**

- Update to use `IxRadioGroup` `@valueChange` event (single handler)
- Correctly extract `event.detail` from `CustomEvent<Breakpoint>`

---

### DIFF
--- a/packages/vue-test-app/src/preview-examples/application-breakpoints.vue
+++ b/packages/vue-test-app/src/preview-examples/application-breakpoints.vue
@@ -75,9 +75,9 @@ export default {
         header-title="Choose breakpoint"
       ></IxContentHeader>
       <IxRadioGroup :value="breakpoints[0]" @valueChange="setBreakpoint">
-        <IxRadio value="sm" label="Small"></IxRadio>
-        <IxRadio value="md" label="Medium"></IxRadio>
-        <IxRadio value="lg" label="Large"></IxRadio>
+        <IxRadio value="sm" label="Small" aria-label="Small"></IxRadio>
+        <IxRadio value="md" label="Medium" aria-label="Medium"></IxRadio>
+        <IxRadio value="lg" label="Large" aria-label="Large"></IxRadio>
       </IxRadioGroup>
     </IxContent>
   </IxApplication>


packages/vue/src/plugin.ts
```diff
@@ -1,24 +1,8 @@
...
-        el.addEventListener(toKebabCase(name), cb, opts);
-      },
-      rel(el: HTMLElement, name, cb, opts) {
-        el.removeEventListener(toKebabCase(name), cb, opts);
-      },
-      ce(name, opts) {
-        return new CustomEvent(toKebabCase(name), opts);
-      },
-    });
+    handlePlatformHelpers({});
```
danielleroux: Do we need the ixPlugin and the platformHelpers in general anymore?

danielleroux (changes requested):

packages/vue/src/plugin.ts
alexkaduk: it is not needed

alexkaduk: now it is like this https://github.com/siemens/ix/pull/2241/files#diff-e2be2a20c4be9bc658a7f7453ef8760ee6edb7d24cf50b307f2d371c4be13d17R5

danielleroux (dismissed):

PR #2245, opened 2025-11-04
refactor(core/radio): add accessibility improvements and host click handler

alexkaduk:


packages/core/src/components/toggle/toggle.tsx
```diff
@@ -178,6 +178,10 @@ export class Toggle implements IxFormComponent<string> {
     }
     return (
       <Host
+        role="switch"
+        aria-label={toggleText}
+        aria-checked={a11yBoolean(this.checked)}
+        aria-disabled={a11yBoolean(this.disabled)}
```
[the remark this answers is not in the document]
alexkaduk: would like to suggest to do it separately as this update makes component to work in diff way it works at the moment. not sure I have handle it correctly at the moment.

packages/core/src/components/checkbox/checkbox.tsx
```diff
@@ -185,6 +185,7 @@ export class Checkbox implements IxFormComponent<string> {
       <Host
         aria-checked={a11yBoolean(this.checked)}
         aria-disabled={a11yBoolean(this.disabled)}
+        aria-label={this.label}
```
danielleroux: aria-label can be removed its not necessary for the accessibility tree
image

packages/core/src/components/toggle/toggle.tsx
```diff
@@ -178,6 +178,10 @@ export class Toggle implements IxFormComponent<string> {
     }
     return (
       <Host
+        role="switch"
+        aria-label={toggleText}
```
danielleroux: aria-label can be removed its not necessary for the accessibility tree
image

packages/core/src/components/checkbox/checkbox.tsx
alexkaduk: removed. thanks

PR #2269, opened 2025-11-16
feat(IX-3657): update styles for date and time picker

alexkaduk:


packages/core/src/components/time-picker/time-picker.tsx
```diff
@@ -409,7 +409,8 @@ export class TimePicker {
     const focusedValue = Number(this.formattedTime[unit]);
 
     this.updateDescriptorFocusedValue(unit, focusedValue);
-    this.elementListScrollToTop(unit, focusedValue, 'smooth');
+    // Removed auto-scroll on blur - it conflicts with manual scrolling
+    // this.elementListScrollToTop(unit, focusedValue, 'smooth');
```
alexkaduk: Removed auto-scroll on blur - it conflicts with manual scrolling

packages/core/src/components/time-picker/time-picker.tsx
```diff
@@ -878,28 +903,35 @@ export class TimePicker {
...
+                    {descriptor.numberArray.map((number) => {
+                      return (
+                        <button
+                          data-element-container-id={`${descriptor.unit}-${number}`}
+                          class={{
+                            selected: this.isSelected(descriptor.unit, number),
+                            'element-container': true,
+                          }}
+                          onMouseDown={(e) => {
+                            e.preventDefault();
```
alexkaduk, line 911: Browser's native "scroll-into-view on focus" behavior was moving partially visible buttons out from under the cursor during the click sequence, causing the wrong button (or no button) to be selected.

packages/core/src/components/date-picker/date-picker.scss
```diff
@@ -14,7 +14,6 @@
 :host {
   display: block;
   position: relative;
-  width: 21rem;
```
lzeiml, line 17: I think we should keep this fixed width for now or it might cause unexpected layout changes for our users.

packages/core/src/components/time-picker/time-picker.scss
```diff
@@ -68,26 +67,29 @@
...
       }
     }
 
     .element-list {
       list-style-type: none;
       overflow: auto;
-      max-height: 15.375rem;
-      padding-top: 0.125rem;
+      padding: 1px;
+      margin: -1px;
```
lzeiml: Please consider using rem instead of px wherever possible. (also check your other changes for this)

packages/core/src/components/time-picker/time-picker.tsx
```diff
@@ -207,6 +207,14 @@ export class TimePicker {
     this._time = timeFormat;
   }
 
+  /**
+   * Get default time value
+   * @returns undefined for empty state (no selection)
+   */
+  private getDefaultTime(): DateTime | undefined {
+    return DateTime.now();
+  }
```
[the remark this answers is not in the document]
alexkaduk: DateTime.now() should be return at the moment

packages/core/src/components/time-picker/time-picker.scss
```diff
@@ -68,26 +67,29 @@
       .column-seperator {
```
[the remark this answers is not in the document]
alexkaduk: fixed

packages/core/src/components/time-picker/time-picker.scss
```diff
@@ -98,27 +100,31 @@
...
 
         &:hover {
           background-color: var(
             --theme-datepicker-day--background--selected-hover
           );
         }
+
+        &:focus-visible {
+          border: 1px solid black;
+        }
```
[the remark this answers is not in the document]
alexkaduk: fixed. thanks

packages/core/src/components/date-picker/date-picker.scss
alexkaduk: we have fixed size for calendar item and for spaces. so it might just do the same as additional selector

packages/core/src/components/time-picker/time-picker.scss
alexkaduk: adjusted. thanks

PR #2275, opened 2025-11-18
fix(link-button/links): add underline for WCAG compliance (IX-3558)

alexkaduk:


lzeiml:
Changeset is missing, other than that looks good!

packages/core/src/components/link-button/link-button.scss
nuke-ellington, line 1: I think we can simplify the hover and active mixins since the underline is set in the default styling already

packages/core/scss/components/_links.scss
```diff
@@ -16,7 +16,8 @@
 
 a[href] {
   color: var(--theme-color-primary);
-  text-decoration: none;
+  text-decoration: underline;
```
nuke-ellington: I think we can remove the decoration from hover and active mixins

packages/core/src/components/link-button/link-button.scss
alexkaduk: done. thanks

packages/core/scss/components/_links.scss
alexkaduk: updated, thanks

PR #2280, opened 2025-11-24
Made card passive

1307-Dev:
Currently card has active and hover states.

JIRA Issue- IX-391

Added a passive property to card, push-card and action-card i.e., it will not get a hover or active state (similar to tile) when passive is set to true.

## 🏁 Checklist

### DIFF
--- a/packages/core/src/components/card/card.scss
+++ b/packages/core/src/components/card/card.scss
@@ -112,15 +112,12 @@
   @include card-variants-selected($variant);
 }
 
-:host(.passive),
-:host(.passive:hover),
-:host(.passive:active) {
-  cursor: default !important;
+:host(.passive) {
+  cursor: default;
 }
 
 @each $variant in card-vars.$ix-card-variants {
-  :host(.passive.card-#{$variant}:hover),
-  :host(.passive.card-#{$variant}:active) {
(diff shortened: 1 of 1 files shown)


packages/core/src/components/card/card.scss
```diff
@@ -111,3 +111,17 @@
   @include card-variants-active($variant);
   @include card-variants-selected($variant);
 }
+
+:host(.passive),
+:host(.passive:hover),
+:host(.passive:active) {
+  cursor: default !important;
+}
```
nuke-ellington: Something like this should be enough

```suggestion
:host(.passive) {
  cursor: default;
}
```

If not we can do something like `:host(:not(.passive)) { cursor: pointer }` instead further above

packages/core/src/components/card/card.scss
```diff
@@ -111,3 +111,17 @@
...
+
+@each $variant in card-vars.$ix-card-variants {
+  :host(.passive.card-#{$variant}:hover),
+  :host(.passive.card-#{$variant}:active) {
```
nuke-ellington: Should also work
```suggestion
  :host(.passive.card-#{$variant}) {
```

Issue #2274, opened 2025-11-18
ix-tree with custom node item does not work in version 4.0


baloghlaszlo99:
Hi!
I updated the siemens/ix library in my project to version 4.0.0 as well I updated angular to 20. I noticed that I face a runtime error in the ix-tree component. The following error message appears in the console: "TypeError: this.renderItem is not a function". This exception pointing to this code part:
`getVirtualizerOptions(refreshTreeOptions) {
        const list = this.buildTreeList(this.model[this.root]);
        return {
            width: "100%",
            height: "100%",
            itemHeight: 32,
            total: list.length,
            generate: index => {
                const item = list[index];
                const renderedTreeItem = this.hostElement.querySelector(`[data-tree-node-id="${item.id}"]`);
                const context = this.getContext(item.id);
                if (renderedTreeItem && refreshTreeOptions.force === false) {
                    renderedTreeItem.hasChildren = item.hasChildren;
                    renderedTreeItem.context = Object.assign({}, context);
                    let forceRerender = this.dirtyItems.has(item.id);
                    if (this.updates.has(item.id)) {
                        const doUpdate = this.updates.get(item.id);
                        if (doUpdate) {
                            const updateRequestedRerender = doUpdate(item, Object.assign({}, this.context));
                            if (typeof updateRequestedRerender === "boolean") {
                                forceRerender = updateRequestedRerender;
                            }
                        }
                    }
                    this.updatePadding(renderedTreeItem, item);
                    if (!forceRerender) {
                        return renderedTreeItem;
                    }
                }
                const update = callback => {
                    this.updates.set(item.id, callback);
                };
                let innerElement = null;
                if (this.renderItem) {
                    innerElement = this.renderItem(index, item, list, Object.assign({}, this.context), update);
                }
                if (innerElement === null) {
                    innerElement = renderDefaultItem(item, context, update);
                }
                const el = innerElement;
                el.setAttribute("data-tree-node-id", item.id);
                el.style.paddingRight = "1rem";
                this.updatePadding(el, item);
                this.dirtyItems.delete(item.id);
                return el;
            }
        };
    }`

Version information:
- Angular: 20.3.12
- Siemens/ix: 4.0.0
- Node: 22.20.0
- Operating system: win 11

I created a Stackblitz example and hope it's available here: 
https://stackblitz.com/edit/sz6df6zc?file=src%2Fapp%2Fapp.module.ts
If you need more detailed information, I'm happy to cooperate and thank you for your help in advance.

[...]

PR #2283, opened 2025-11-24
fix(angular/tree): remove renderItem from ProxyCmp inputs to enable custom rendering #IX-3668

alexkaduk:

ix-tree is not rendered for angular

GitHub Issue Number: #2274

Fixed an issue where **ix-tree** with a custom `renderItem` function would throw a runtime error ("TypeError: this.renderItem is not a function")

## 🏁 Checklist

### DIFF
--- a/.changeset/few-cars-shake.md
+++ b/.changeset/few-cars-shake.md
@@ -1,0 +1,7 @@
+---
+'@siemens/ix-angular': patch
+---
+
+An issue for `ModalService` and `ToastService` in **ix-angular/standalone** has been fixed and can now be injected correctly.
+
+Fixes #2060
--- a/.changeset/fix-invalidtext-property.md
+++ b/.changeset/fix-invalidtext-property.md
@@ -1,0 +1,9 @@
+---
+'@siemens/ix': patch
+---
+
+Fix `invalidText` property not being applied to **ix-date-input** and **ix-time-input** when internal validation fails.
+
+The `invalidText` property now correctly takes precedence over the i18n error messages when both are set. Previously, the i18n message would always be shown for internal validation errors (unparsable dates/times or min/max violations), ignoring the user's custom `invalidText`.
+
+Fixes #2183.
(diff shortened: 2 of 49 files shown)


.changeset/silent-tires-fall.md
```diff
@@ -0,0 +1,5 @@
+---
+'@siemens/ix-angular': patch
+---
+
+remove renderItem from ProxyCmp inputs to enable custom rendering
```
lzeiml:
```suggestion
Fixed an issue where **ix-tree** with a custom `renderItem` function would throw a runtime error ("TypeError: this.renderItem is not a function")

Fixes #2274
```

lzeiml (approved):

packages/angular/src/tree.ts
```diff
@@ -14,11 +14,11 @@ import {
   ElementRef,
   NgZone,
 } from '@angular/core';
-import { ProxyCmp, proxyOutputs } from './angular-component-lib/utils';
 import { TreeBaseDirective } from '@siemens/ix-angular/common';
+import { ProxyCmp, proxyOutputs } from './angular-component-lib/utils';
 
 @ProxyCmp({
-  inputs: ['context', 'model', 'root', 'renderItem'],
```
nuke-ellington, line 21: Wouldn't it be better to check for the definition of a member in the proxy component? This way optional properties will still get picked up if set.

alexkaduk: ProxyCmp here

https://github.com/siemens/ix/blob/441e314b70408157c26e6c1f082e3f6f7d487607/packages/angular/src/angular-component-lib/utils.ts#L13

can't detect that TreeBaseDirective 

https://github.com/siemens/ix/blob/441e314b70408157c26e6c1f082e3f6f7d487607/packages/angular/common/src/directives/tree.ts#L47

already handles renderItem with a custom setter. 

Including it in both places creates a conflict. Manual exclusion is needed to specify "this property is handled by the base class, don't proxy it."

In case I do understand your comment incorrectly, pls clarify a bit.

PR #2300, opened 2025-12-07
fix(html-test-app): apply theme from URL params in preview iframes #IX-3736

alexkaduk:

eCharts examples in preview iframes always display with classic theme colors, even when ?theme=brand&colorSchema=dark URL parameters are present.

**Root Cause:**
- init.js loaded after chart initialization
- Charts called themeSwitcher.getCurrentTheme() before data attributes were set
- Hardcoded class="theme-classic-dark" on body tags interfered with theme detection 

Fixed all 15 ECharts HTML files:
- Removed hardcoded theme-classic-dark class from <body> tags
- Moved <script type="module" src="./init.js"></script> to load before chart initialization

**Result:**
- Data attributes are set before charts initialize
- themeSwitcher.getCurrentTheme() returns correct theme
- Charts properly render with correct theme and switch between light/dark modes ✅

## 🏁 Checklist

### DIFF
--- a/packages/html-test-app/src/preview-examples/init.js
+++ b/packages/html-test-app/src/preview-examples/init.js
@@ -149,7 +149,6 @@ function setBodySizes() {
   document.head.appendChild(styleElement);
 }
 
-// Initialize and expose promise for chart scripts to await
 async function init() {
   ixIconsDefineCustomElements();
   defineCustomElements();
@@ -167,4 +166,5 @@ async function init() {
   header.appendChild(scrollbarStyle);
 }
 
+// Initialize and expose promise for chart scripts to await
 globalThis.ixInitPromise = init();


.changeset/preview-theme-url-params.md
lzeiml, line 1: There is no changeset need for this PR because we do not release a changelog for the test-apps.

lzeiml (changes requested):

.changeset/preview-theme-url-params.md
alexkaduk: removed. thanks

lzeiml (dismissed):

packages/html-test-app/src/preview-examples/init.js
```diff
@@ -64,15 +65,18 @@ function loadAdditionalTheme() {
...
-  }
+  const fullTheme = `theme-${theme}-${colorSchema}`;
+
+  themeSwitcher.setTheme(fullTheme);
+
+  const themeClasses = Array.from(document.body.classList).filter((cls) =>
+    cls.startsWith('theme-')
+  );
+  document.body.classList.remove(...themeClasses);
+  document.body.classList.add(fullTheme);
```
nuke-ellington: Is this required? Should already be handled by the `replaceThemeClass` method of the theme switcher.

alexkaduk: removed. thanks!

could you pls create new snapshot, it can be checked with docu?

packages/echarts/src/computecss.ts
```diff
@@ -7,7 +7,8 @@
  * LICENSE file in the root directory of this source tree.
  */
 
-
 export default function getComputedCSSProperty(cssProperty: string) {
-  return getComputedStyle(document.body).getPropertyValue(`--theme-${cssProperty}`);
+  return getComputedStyle(document.documentElement).getPropertyValue(
```
nuke-ellington: IMO this will break backward comp for this use case: <body class="theme-..
Maybe we can check both?

.changeset/strange-needles-relate.md
nuke-ellington, line 1: No changeset for test apps needed.

nuke-ellington (changes requested):

packages/echarts/src/computecss.ts
alexkaduk: no changes here

.changeset/strange-needles-relate.md
alexkaduk: moved. thanks

PR #2309, opened 2025-12-11
fix(core/chip, pill): removed default left margin from chip pill

GayatriK2002:

- Chip and pill has default left margin

Jira Issue Number: 3178

- Removed chip and pill default left margin

## 🏁 Checklist


testing/visual-testing/tests/blind/custom-header/index.html
nuke-ellington, line 1: This example should not have the margin (this will update the snapshot)

nuke-ellington (changes requested):

testing/visual-testing/tests/blind/custom-header/index.html
GayatriK2002: The blind/custom-header VRT contains ix-pill components, so I added the left margin in the custom-header VRT. Without it, the VRT was failing.

testing/visual-testing/tests/pill/basic/index.html
```diff
@@ -13,6 +13,11 @@
       content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0"
     />
     <title>Stencil Component Starter</title>
+    <style>
+      .pill-left-margin {
+        margin-left: 0.25rem;
+      }
+    </style>
```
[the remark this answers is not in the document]
GayatriK2002: I initially tried using `display: flex` with `gap: 0.25rem` as you suggested, which is indeed cleaner. However, it caused visual regression test failures because the spacing is applied differently than the original `margin-left` that was defined in the component's Shadow DOM.

The original implementation had `margin-left: 0.25rem` on the `:host` selector (applied to every chip/pill instance), and using flexbox `gap` creates subtle pixel-level differences in positioning that don't match the baseline snapshots. To maintain backward compatibility and avoid regenerating all VRT baselines, I've kept the approach that most closely mimics the original behavior.

testing/visual-testing/tests/pill/basic/index.html
```diff
@@ -23,118 +28,107 @@
...
         icon="print"
         style="width: 10rem"
         onClose="{console.log}"
         outline
+        class="pill-left-margin"
+        >Custom</ix-pill
       >
-        Custom
-      </ix-pill>
     </div>
```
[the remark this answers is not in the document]
GayatriK2002: I initially tried using `display: flex` with `gap: 0.25rem` as you suggested, which is indeed cleaner. However, it caused visual regression test failures because the spacing is applied differently than the original `margin-left` that was defined in the component's Shadow DOM.

The original implementation had `margin-left: 0.25rem` on the `:host` selector (applied to every chip/pill instance), and using flexbox `gap` creates subtle pixel-level differences in positioning that don't match the baseline snapshots. To maintain backward compatibility and avoid regenerating all VRT baselines, I've kept the approach that most closely mimics the original behavior.

testing/visual-testing/tests/blind/custom-header/index.html
nuke-ellington: It should fail in this case since the margin is not needed anymore. Just update the images for this test case after removing the margins 👍

testing/visual-testing/tests/chip/basic/index.html
```diff
@@ -13,6 +13,11 @@
       content="width=device-width, initial-scale=1.0, minimum-scale=1.0, maximum-scale=5.0"
     />
     <title>Stencil Component Starter</title>
+    <style>
+      .chip-left-margin {
```
nuke-ellington: Can we not simplify the selector? This way we won't have to add classes to all elements (also applies to other test files).
```suggestion
      ix-chip {
```

nuke-ellington (changes requested):

testing/visual-testing/tests/blind/custom-header/index.html
GayatriK2002: Yes, updated the images

testing/visual-testing/tests/chip/basic/index.html
GayatriK2002: Updated the code

Issue #2190, opened 2025-10-09
Modal without ix-* element at the top of template


piotretm:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

After upgrading @siemens/ix from 3.1.0 to 3.2.0 we've discovered an issue with modals. Generally, if I don't put an `ix-*` element (it does not have to be modal related element, it can simply be an `ix-icon-button` for instance) as the very first in the template it is not working properly.

If I try to call `showModal` with `centered` it is only centered on the first interaction. Later it is not.

I am providing an [example](https://stackblitz.com/edit/hk8astif?file=src%2Findex.html) from your page with small modifications, so you can see what is the problem. Please try to remove the `ix-button` from the modal template and open it multiple times.

If you remove the button, first of all it is not rendering any content. Also, the first modal is opened with styles:

`{
    opacity: 1;
    transform: translateY(-50%) translateX(-50%);
}`
 but next ones always have a different styling:
`{
    opacity: 1;
    transform: translateY(40px) translateX(-50%);
}`

### What type of frontend framework are you seeing the problem on?

JavaScript

### On which version of the frontend framework are you experiencing the issue?

lit@3.3.1

### Which version of iX do you use?

3,2,0

### Code to produce this issue.

```shell
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>Modal sizes example</title>
    <link rel="stylesheet" href="./modal-sizes.css" />
  </head>
  <body>
    <div class="modal-sizes">
      <ix-button data-modal-size="360">Show modal size 360</ix-button>
    </div>

    <template id="modal-example-template">
      <ix-button>CLICK</ix-button>
      <div>hi</div>
    </template>

    <script type="module">
      import { showModal, dismissModal } from '@siemens/ix';

      function createExampleModal() {
        const name = 'modal-example';
        window.customElements.define(
          name,
          class extends HTMLElement {
            isInitalRender = false;

            constructor() {
              super();

              this.size = '360';
            }

            connectedCallback() {
              if (this.isInitalRender) {
                return;
              }

              this.isInitalRender = true;
              this.firstRender();
            }

            firstRender() {
              const modalTemplate = document.getElementById(
[...]

PR #2329, opened 2026-01-07
fix(core/modal): Modal not centered and content not rendered

RamVinayMandal:

When using showModal with centered: true option, the modal only centers correctly on the first open. If the modal is dismissed and reopened, it fails to center properly because the slideInModal() animation method was only called during componentDidLoad() lifecycle hook, which only fires once when the component is first mounted.

> The modal content appearing empty is not caused by the ix component.
>  
> The issue occurs when the Cancel/OK button is removed. This causes the related event listeners to become null, which leads to a browser error.
>  
> No additional checks were added in the example in order to keep it simple and consistent with the original implementation.

GitHub Issue Number: #2190

JIRA ID: [IX-3455](https://agileworld.siemens.cloud/jira/browse/IX-3455)

Additionally covered: JIRA ID: [IX-3808](https://agileworld.siemens.cloud/jira/browse/IX-3808)
Please check the comment section also of this JIRA ticket. A minor styling fix is added for long string without space which overflows beyond the modal.

- Modal now correctly centers on every open, including subsequent reopens after being dismissed
- Moved slideInModal() call from componentDidLoad() to showModal() method, ensuring the centering animation runs each time the modal is opened
- Added test case to verify centered modals remain centered after reopening

Note: The content rendering issue mentioned in the original issue (content not rendered when there's no ix-* component at the top of the template) was not reproducible during investigation. The modal content renders correctly regardless of whether an ix-* component is the first element.

## 🏁 Checklist

### DIFF
--- a/.changeset/tender-games-drive.md
+++ b/.changeset/tender-games-drive.md
@@ -1,0 +1,7 @@
+---
+"@siemens/ix": patch
+---
+
+Make sure centered **ix-modal** will get displayed in desired place on every render.
+
+Fixes #2190


packages/core/src/components/modal/test/modal.ct.ts
```diff
@@ -337,3 +337,44 @@ regressionTest.describe('message utils', () => {
...
+      page.evaluate(() => {
+        const elm = document.createElement('ix-modal');
+        elm.innerHTML = `<div>hi</div>`;
+        window.showModal({ content: elm, centered: true });
+      });
+
+    const getModalCenterX = async () => {
+      const box = await dialog.boundingBox();
+      return box!.x + box!.width / 2;
+    };
```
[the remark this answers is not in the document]
RamVinayMandal: Accepted the suggested change

packages/core/src/components/modal/modal.scss
```diff
@@ -46,6 +46,7 @@
     overflow: visible;
     max-height: 80vh;
     pointer-events: all;
+    word-wrap: break-word;
```
nuke-ellington: We should try to handle the overflow by placing words that don't fit on the next line instead of breaking them, since this should make the text hard to read.

nuke-ellington:
```suggestion
    overflow-wrap: break-word;
```

PR #2341, opened 2026-01-15
feat(dropdown): add enableTopLayer prop for Popover API rendering #IX-3787

alexkaduk:

Dropdowns in `ix-dropdown` and consuming components (dropdown-button, select, etc.) can be clipped or hidden behind elements with higher z-index values, particularly in AG Grid cells. This is caused by CSS stacking context limitations where no z-index value can escape the parent's stacking context.

**Example Issue:**
- Dropdown-button in AG Grid cells appears **behind** subsequent rows
- Select components get cut off by grid content
- Impossible to display dropdowns above complex components like AG Grid

Added an **opt-in** `enableTopLayer` prop to `ix-dropdown` and all 11 consuming components that enables rendering in the browser's top layer using the Popover API.

**Key Changes:**

✅ **New `enableTopLayer` prop** (default: `false`)
- When `true`: Renders dropdown using `<dialog popover="manual">` in browser's top layer
- When `false`: Uses legacy fixed positioning (backward compatible)

✅ **Updated 11 Components:**
- `ix-dropdown-button`, `ix-split-button`, `ix-select`
- `ix-date-input`, `ix-time-input`, `ix-date-dropdown`, `ix-date-picker`
- `ix-category-filter`, `ix-application-header`, `ix-breadcrumb`, `ix-menu-avatar`

### DIFF
--- a/packages/core/src/components/dropdown/dropdown.tsx
+++ b/packages/core/src/components/dropdown/dropdown.tsx
@@ -633,10 +633,10 @@ export class Dropdown implements ComponentInterface, DropdownInterface {
               overflow: !this.suppressOverflowBehavior,
             }}
             popover="manual"
-            aria-modal="false"
+            aria-modal="true"
+            aria-label={this.header}
             tabindex={-1}
             onClick={(event: PointerEvent) => this.onDropdownClick(event)}
-            onKeyDown={() => {}}
           >
             <div class="dropdown-container">
               {this.header && <div class="dropdown-header">{this.header}</div>}
--- a/packages/core/src/components/dropdown/test/dropdown-top-layer.ct.ts
+++ b/packages/core/src/components/dropdown/test/dropdown-top-layer.ct.ts
@@ -16,16 +16,18 @@ regressionTest.describe('enableTopLayer feature', () => {
       async ({ mount, page }) => {
         await mount(`
         <ix-button id="trigger">Open</ix-button>
-        <ix-dropdown id="dropdown" trigger="trigger" enable-top-layer="true">
+        <ix-dropdown id="dropdown" trigger="trigger" enable-top-layer="true" header="Menu">
           <ix-dropdown-item label="Item 1"></ix-dropdown-item>
           <ix-dropdown-item label="Item 2"></ix-dropdown-item>
         </ix-dropdown>
       `);
 
-        const dropdown = page.locator('ix-dropdown#dropdown');
-
-        const dialog = dropdown.locator('dialog[popover="manual"]');
+        const dialog = page.locator(
+          'ix-dropdown#dropdown dialog[popover="manual"]'
+        );
         await expect(dialog).toBeAttached();
+        await expect(dialog).toHaveAttribute('aria-modal', 'true');
+        await expect(dialog).toHaveAttribute('aria-label', 'Menu');
       }
     );
 
@@ -34,15 +36,14 @@ regressionTest.describe('enableTopLayer feature', () => {
       async ({ mount, page }) => {
         await mount(`
         <ix-button id="trigger">Open</ix-button>
-        <ix-dropdown id="dropdown" trigger="trigger" enable-top-layer="true">
+        <ix-dropdown id="dropdown" trigger="trigger" enable-top-layer="true" header="Menu">
           <ix-dropdown-item label="Item 1"></ix-dropdown-item>
           <ix-dropdown-item label="Item 2"></ix-dropdown-item>
         </ix-dropdown>
       `);
 
-        const trigger = page.locator('#trigger');
-        const dropdown = page.locator('ix-dropdown#dropdown');
-        const dialog = dropdown.locator('dialog[popover="manual"]');
+        const trigger = page.getByRole('button', { name: 'Open' });
+        const dialog = page.getByRole('dialog');
 
         await expect(dialog).not.toBeVisible();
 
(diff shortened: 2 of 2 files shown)


packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -542,27 +615,57 @@ export class Dropdown implements ComponentInterface, DropdownInterface {
...
+            onKeyDown={() => {}}
+          >
+            <div class="dropdown-container">
+              {this.header && <div class="dropdown-header">{this.header}</div>}
+              {this.show && <slot></slot>}
+            </div>
+          </dialog>
+        </Host>
+      );
+    } else {
```
danielleroux: Dont need to provide else branch here.
```tsx
if (this.enableTopLayer) {
    return (<Host>....</Host>)
}


return (<Host>...</Host>);

```

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -404,17 +419,76 @@ export class Dropdown implements ComponentInterface, DropdownInterface {
...
+    }
+
+    popover.showPopover();
+
+    await this.resolveAnchorElement();
+
+    if (this.anchorElement) {
+      this.applyDropdownPosition();
+    } else {
+      requestAnimationFrameNoNgZone(() => {
```
danielleroux, line 437: Why is it necessary to wait for an animation frame?

packages/core/src/components/dropdown/test/dropdown-top-layer.ct.ts
danielleroux, line 1: Would be nice to test the benefits of having the top-layer. e.g overflow behaviour
```
<div style="overflow: hidden; position: absolute">
    <button>
    </button>
    <ix-dropdown trigger=".." enabel-top-layer>
         <ix-dropdown-items....
```

Ensure that the dropdown is fully visible and not hidden inside the div

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -374,11 +378,22 @@ export class Dropdown implements ComponentInterface, DropdownInterface {
       });
 
       this.registerKeyListener();
+
+      if (this.enableTopLayer) {
```
danielleroux: Regarding more readable code we could refactor that stuff an to a early return instead of if else branches nesting

packages/core/src/components/dropdown/dropdown.tsx
alexkaduk: updated. thanks

packages/core/src/components/dropdown/dropdown.tsx
alexkaduk: same as line 549 `focusDropdownItem`
when did migration to popover API, some tests failed. adding requestAnimationFrameNoNgZone resolved the case. should it be removed?

packages/core/src/components/dropdown/dropdown.tsx
alexkaduk: updated changedShow. thanks
applied early return

packages/core/src/components/dropdown/test/dropdown-top-layer.ct.ts
alexkaduk: AG grid has such
https://github.com/siemens/ix/pull/2341/files#diff-617d4eb1a0ed48025dfb90ad5770595ed587dffe200e3525218f9169df7bf039R125
also added case `escapes CSS stacking context - dropdown is fully visible and not hidden (enableTopLayer=true)` to `dropdown-top-layer.ct.ts`

packages/core/src/components/dropdown/dropdown.tsx
danielleroux: So you think its more related to the async behaviour of the dialog api?

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -542,6 +617,36 @@ export class Dropdown implements ComponentInterface, DropdownInterface {
...
+          }}
+        >
+          <dialog
+            ref={this.dialogRef}
+            class={{
+              dialog: true,
+              overflow: !this.suppressOverflowBehavior,
+            }}
+            popover="manual"
+            aria-modal="false"
```
danielleroux: aria-modal should be `true` and would also make sense to provide additional aria labels here: https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/#wai-ariaroles,states,andproperties

packages/core/src/components/dropdown/dropdown.tsx
```diff
@@ -542,6 +617,36 @@ export class Dropdown implements ComponentInterface, DropdownInterface {
...
+            tabindex={-1}
+            onClick={(event: PointerEvent) => this.onDropdownClick(event)}
+            onKeyDown={() => {}}
```
danielleroux: Empty listener not used can be removed

packages/core/src/components/dropdown/test/dropdown-top-layer.ct.ts
danielleroux, line 1: After introducing aria-labels we should not use locator function based on css selectors, we should use the accessibility locator functions (https://playwright.dev/docs/locators) this ensure a better accessibility coverage, and avoid breaking changes if the change structure

danielleroux (changes requested):

packages/core/src/components/dropdown/dropdown.tsx
alexkaduk: yes. might be it

packages/core/src/components/dropdown/dropdown.tsx
alexkaduk: removed. thanks
after I have +1 sonar issue. JFYI

packages/core/src/components/dropdown/test/dropdown-top-layer.ct.ts
alexkaduk: applied. thanks

PR #2352, opened 2026-01-21
test: integrate framework smoke tests

danielleroux:

Add for each preview example inside the test-apps, a small testcase that ensure that the component is rendered correctly. I choose aria snapshots to find or identify missing aria definitions and also to relay on a more stable snapshot instead of using the dom snapshot.

Also added some route types to ensure that the preview apps are defining all the same routes:

Implement route types:

- [x] angular
- [x] angular standalone
- [x] react
- [x] vue (Currently not working because optional properties which are not undefined on purpose will be rendered with the string value `Symbol()`) https://github.com/stenciljs/output-targets/blob/61cd6ada5915e2331f9eaffd1bd855a6be3a8752/packages/vue/src/runtime.ts#L33

All other test-apps will be ignored because there is no dedicated wrapper library behind

Following PR is **required** https://github.com/siemens/ix-icons/pull/112 this ensures that the fallback icon name is even given if the icon is provided via import like `iconRocket`

## 🏁 Checklist


testing/framework-tests/package.json
```diff
@@ -0,0 +1,26 @@
...
+  "name": "framework-tests",
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "build": "tsx src/main.ts",
+    "test.setup": "playwright install chromium --with-deps"
+  },
+  "devDependencies": {
+    "@axe-core/playwright": "^4.11.0",
```
nuke-ellington: Add to catalog?

danielleroux: Good point!

nuke-ellington (approved):

Issue #2326, opened 2026-01-06
Application menu - Navigation menu option 'enableToggleTheme' does not toggle the theme


benvla:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

The [navigation menu option enableToggleTheme](https://ix.siemens.io/docs/components/application-menu/guide#options) of the Application menu correctly shows the toggle theme button, however it does not change the theme.

This option works in v3, it doesn't work in v4.

### What type of frontend framework are you seeing the problem on?

Angular (Standalone)

### On which version of the frontend framework are you experiencing the issue?

@angular/core@20.3.15

### Which version of iX do you use?

v4.2.0

### Code to produce this issue.

```shell
Failing on iX version 4:
https://stackblitz.com/edit/s9m4y9w6?file=src%2Fvertical-tabs.ts

Working example on iX version 3:
https://stackblitz.com/edit/vixgcace?file=src%2Fvertical-tabs.ts
```

PR #2357, opened 2026-01-23
fix(core/themeswitcher) : enableToggleTheme does not toggle theme in menu component

lakshmi-priya-b:

- When we try to toggle theme using 'enableToggleTheme' property in menu-component it is not working.
- 'enableToggleTheme' is internally calling 'themeswitcher'.
- Themeswitcher was modified to always target html element but selectors was still supporting only the body elements.

GitHub Issue Number: #2326 

- When we try to toggle theme using 'enableToggleTheme' property in menu-component it is  working.
- Added selectors for html element as well.

## 🏁 Checklist

### DIFF
--- a/packages/core/src/components/utils/test/theme-switcher.spec.ts
+++ b/packages/core/src/components/utils/test/theme-switcher.spec.ts
@@ -77,11 +77,11 @@ describe('ThemeSwitcher', () => {
   });
   describe('getMode', () => {
     it('should return dark when data-ix-color-schema is dark', () => {
-      document.documentElement.setAttribute('data-ix-color-schema', 'dark');
+      document.documentElement.dataset.ixColorSchema = 'dark';
       expect(themeSwitcher.getMode()).toBe('dark');
     });
     it('should return light when data-ix-color-schema is light', () => {
-      document.documentElement.setAttribute('data-ix-color-schema', 'light');
+      document.documentElement.dataset.ixColorSchema = 'light';
       expect(themeSwitcher.getMode()).toBe('light');
     });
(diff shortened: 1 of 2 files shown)


lakshmi-priya-b:
Updated CSS class-based theme deprecation from v4.0.0 to v5.0.0 in the ix-docs.
PR-link : https://github.com/siemens/ix-docs/pull/176

packages/core/scss/theme/classic/theme.scss
nuke-ellington, line 1: Refactoring the selector to target the HTML element will introduce other problems. Instead fix theme switcher itself.

packages/core/src/components/utils/test/theme-switcher.spec.ts
nuke-ellington, line 1: The test claim to test changing CSS classes but then test data attributes instead.

nuke-ellington (changes requested):
We should not map CSS classes to data attributes to maintain backward compatibility. Instead we could introduce a getMode fn that can then be called theme imlementation agnostic by e.g the menu component.

packages/core/src/components/utils/test/theme-switcher.spec.ts
lakshmi-priya-b: Updated👍

packages/core/src/components/utils/theme-switcher.ts
nuke-ellington, line 1: Since setting CSS theme classes on the HTML element is not supported (and never was) we only need to check body for class based theme switching.

PR #2361, opened 2026-01-27
fix(core): Updated aria labels for ix-icon-buttons

GayatriK2002:

- Icon-only buttons announce as "unknown" to screen readers because aria-label attributes are removed after first render and lost on component re-renders.

Jira Issue Number: 3846, 3845

- Added persistent ariaLabelButton prop to ix-icon-button with context-aware labels (e.g., "Collapse pane"/"Expand pane", "Open calendar", "Clear selection") that survive re-renders. Fallback to icon name parsing if no label provided. Users can override via props or host attributes.

## 🏁 Checklist

## 👨‍💻 Help & support

### DIFF
--- a/packages/core/src/components.d.ts
+++ b/packages/core/src/components.d.ts
@@ -2765,15 +2765,16 @@ export namespace Components {
          */
         "i18nPage": string;
         /**
-          * Number of items shown at once. Can only be changed in advaced mode.
+          * Number of items shown at once. Can only be changed in advanced mode.
           * @default 15
          */
         "itemCount": number;
         /**
-          * Custom item count options for advanced mode. Provide an array of numbers to display in the items per page dropdown. If not provided or empty, defaults to [10, 15, 20, 40, 100]. Only positive integers greater than 0 are valid. Invalid values and duplicates are automatically filtered out.
-          * @since 4.3.0
+          * Custom item count options for advanced mode. Provide an array of positive numbers to display in the items per page dropdown.
+          * @since 4.4.0
+          * @default [10, 15, 20, 40, 100]
          */
-        "itemCountOptions"?: number[];
+        "itemCountOptions": number[];
         /**
           * Zero based index of currently selected page
           * @default 0
@@ -8925,13 +8926,14 @@ declare namespace LocalJSX {
          */
         "i18nPage"?: string;
         /**
-          * Number of items shown at once. Can only be changed in advaced mode.
+          * Number of items shown at once. Can only be changed in advanced mode.
           * @default 15
          */
         "itemCount"?: number;
         /**
-          * Custom item count options for advanced mode. Provide an array of numbers to display in the items per page dropdown. If not provided or empty, defaults to [10, 15, 20, 40, 100]. Only positive integers greater than 0 are valid. Invalid values and duplicates are automatically filtered out.
-          * @since 4.3.0
+          * Custom item count options for advanced mode. Provide an array of positive numbers to display in the items per page dropdown.
+          * @since 4.4.0
+          * @default [10, 15, 20, 40, 100]
          */
         "itemCountOptions"?: number[];
         /**
--- a/packages/core/src/components/pagination/pagination.tsx
+++ b/packages/core/src/components/pagination/pagination.tsx
@@ -15,7 +15,6 @@ import {
   h,
   Host,
   Prop,
-  Watch,
 } from '@stencil/core';
 import { BaseButton, BaseButtonProps } from '../button/base-button';
 import { a11yBoolean } from '../utils/a11y';
@@ -42,8 +41,6 @@ export class Pagination {
   };
 
   private readonly maxCountPages = 7;
-  private readonly defaultItemCountOptions = [10, 15, 20, 40, 100];
-  private verifiedItemCountOptions?: number[];
 
   @Element() hostElement!: HTMLIxPaginationElement;
 
@@ -54,7 +51,7 @@ export class Pagination {
 
   /**
    * Number of items shown at once.
-   * Can only be changed in advaced mode.
+   * Can only be changed in advanced mode.
    */
   @Prop() itemCount = 15;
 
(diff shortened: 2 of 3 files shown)


packages/core/src/components/icon-button/icon-button.tsx
nuke-ellington, line 1: Why is this change necessary? We can still set the aria-label on the button.

packages/core/src/components/dropdown-button/dropdown-button.tsx
```diff
@@ -110,7 +110,9 @@ export class DropdownButton {
               variant={this.variant}
               disabled={this.disabled}
               alignment="start"
-              ariaLabel={this.ariaLabelDropdownButton}
+              ariaLabelButton={
```
nuke-ellington, line 113: This change renders the `ariaLabelDropdownButton` useless.

packages/core/src/components/expanding-search/expanding-search.tsx
```diff
@@ -183,7 +183,7 @@ export class ExpandingSearch
             onFocus={() => (this.hasFocus = true)}
             onInput={(e: InputEvent) => this.onChange(e)}
             tabindex={this.expanded ? 0 : -1}
-            aria-label={this.ariaLabelSearchInput}
```
nuke-ellington, line 186: `aria-label` was correct

packages/core/src/components/menu/menu-expand-icon.tsx
```diff
@@ -67,6 +67,7 @@ export class MenuExpandIcon {
         tabindex={-1}
         icon={this.expanded ? iconDoubleChevronLeft : iconDoubleChevronRight}
         variant="subtle-tertiary"
+        ariaLabelButton={this.expanded ? 'Collapse menu' : 'Expand menu'}
```
nuke-ellington:
```suggestion
        aria-label={this.expanded ? 'Collapse menu' : 'Expand menu'}
```

packages/core/src/components/pane/pane.tsx
```diff
@@ -118,7 +118,7 @@ export class Pane {
   /**
    * ARIA label close or collapse button
    */
-  @Prop() ariaLabelCollapseCloseButton?: string;
+  @Prop() ariaLabelCollapseCloseButton?: string = 'Toggle pane';
```
nuke-ellington: Why is the default different from the string used during interaction?

packages/core/src/components/tree-item/tree-item.tsx
```diff
@@ -35,7 +35,7 @@ export class TreeItem {
   /**
    * ARIA label for the chevron icon
    */
-  @Prop() ariaLabelChevronIcon?: string;
+  @Prop() ariaLabelChevronIcon?: string = 'Toggle tree item';
```
nuke-ellington: Has no effect anymore

nuke-ellington (changes requested):

packages/core/src/components/icon-button/icon-button.tsx
GayatriK2002: Yes, removed the added property and updated ariaLabel with aria-Label for all icon-buttons used in code

packages/core/src/components/dropdown-button/dropdown-button.tsx
GayatriK2002: updated code accordingly

packages/core/src/components/expanding-search/expanding-search.tsx
GayatriK2002: updated

packages/core/src/components/pane/pane.tsx
GayatriK2002: updated the code accordingly

packages/core/src/components/tree-item/tree-item.tsx
GayatriK2002: updated the code accordingly

packages/core/src/components/date-dropdown/date-dropdown.tsx
```diff
@@ -355,7 +355,7 @@ export class DateDropdown
           icon={iconHistory}
           ref={this.triggerRef}
           disabled={this.disabled}
-          ariaLabel={this.ariaLabelDropdownButton}
+          ariaLabelButton={this.ariaLabelDropdownButton}
```
nuke-ellington, line 358: does not exist anymore?

packages/core/src/components/dropdown-button/dropdown-button.tsx
nuke-ellington: ariaLabelButton does not exist anymore?

GayatriK2002: Removed 'ariaLabelButton' property from ix-icon-button as it uses standard aria-label attribute on the host element. ix-button needs 'ariaLabelButton' because it can have text content separate from it's aria-label whereas ix-icon-button has only an icon so the standard aria-label on host is sufficient. Also using 'ariaLabelChevronDownIconButton', 'ariaLabelDropdownButton', etc. already defined property label in corresponding components so explicit ariaLabelButton was not required in ix-icon-button

packages/core/src/components/date-dropdown/date-dropdown.tsx
GayatriK2002: Changed to ariaLabelButton because these are [ix-button] components, which use the ariaLabelButton property. The [ariaLabelButton] property exists in ix-button

packages/core/src/components/expanding-search/expanding-search.tsx
```diff
@@ -150,7 +150,7 @@ export class ExpandingSearch
             'btn-search': true,
             'btn-search--expanded': this.expanded,
           }}
-          aria-label={this.ariaLabelSearchIconButton}
+          aria-label={this.expanded ? 'Close search' : 'Open search'}
```
nuke-ellington: If you don't need `ariaLabelSearchIconButton` anymore, we can deprecate it but than still use it if set by user

GayatriK2002: ariaLabelSearchIconButton now falls back to dynamic labels ('Open search' / 'Close search') when not set by user. No deprecation needed.

packages/core/src/components/pane/pane.tsx
```diff
@@ -651,6 +662,7 @@ export class Pane {
       >
         <aside
           id={`pane-${this.composition}`}
+          role="none"
```
nuke-ellington:
```suggestion
```

GayatriK2002: Added role="none" to suppress the unnamed complementary landmark that screen readers were announcing before the close button.

nuke-ellington: Have you tried something like this?

`role="region" aria-label={this.heading}`

GayatriK2002: Yes, working on the same

nuke-ellington: Please create separate PR for that. I will merge this one now :)

PR #2405, opened 2026-02-17
feat(toggle): improve accessibility #IX-3949

alexkaduk:

`ix-toggle` did not consistently expose **switch** semantics with a **stable** accessible name separate from visible on/off copy; tests and examples often relied on visible text or internal `input` for assertions.

- `role="switch"` with `aria-checked` (`true` / `false` / `mixed` when indeterminate), `aria-disabled` / `aria-required` as applicable.
- Accessible name from host `aria-label` or `aria-labelledby` only. Visible on/off (and mixed) text is `aria-hidden` and is not used as the name; state stays in `aria-checked`.
- Keyboard: Space and Enter toggle when not disabled.
- `aria-label` follows `InheritAriaAttributesMixin` so dynamic host updates stay in sync; when `aria-labelledby` is present, it takes precedence for naming.

Tests & alignment

- Playwright CT (`toggle.ct.ts`) updated for host `aria-*` and `getByRole('switch', { name })` with explicit `aria-label`.
- Storybook: `a11y-toggle.stories.ts` (strict axe) documents required host naming (including `hide-text`).
- Preview examples (HTML / React / Vue / Angular / standalone): consistent `aria-label` on toggles.
- Framework smoke aria snapshots and visual toggle fixtures updated where needed.

## 🏁 Checklist

### DIFF
--- a/.changeset/toggle-accessibility-improvements.md
+++ b/.changeset/toggle-accessibility-improvements.md
@@ -1,0 +1,5 @@
+---
+'@siemens/ix': patch
+---
+
+Improve **ix-toggle** accessibility (switch semantics, keyboard support, ARIA) and add automated accessibility coverage.
--- a/.changeset/toggle-accessibility-improvements.md
+++ b/.changeset/toggle-accessibility-improvements.md
@@ -2,4 +2,4 @@
 '@siemens/ix': patch
 ---
 
-**ix-toggle:** Exposes correct **switch** semantics for assistive technologies (`role`, `aria-checked`, `aria-disabled`), supports expected **keyboard** interaction, and applies **`aria-label` / `aria-labelledby` and related host ARIA** on the control so screen readers get an accurate name.
+**ix-toggle:** Exposes correct **switch** semantics (`role`, `aria-checked`, `aria-disabled`), keyboard interaction (**Space** / **Enter**), and a **stable accessible name** aligned with the [APG switch pattern](https://www.w3.org/WAI/ARIA/apg/patterns/switch/) (set **`aria-label`** or **`aria-labelledby`** on the host; name does not change with state; state is conveyed via `aria-checked`). Visible on/off copy is **not** used as the accessible name.
--- a/.changeset/toggle-accessibility-improvements.md
+++ b/.changeset/toggle-accessibility-improvements.md
@@ -2,4 +2,4 @@
 '@siemens/ix': patch
 ---
 
-**ix-toggle:** Exposes correct **switch** semantics (`role`, `aria-checked`, `aria-disabled`), keyboard interaction (**Space** / **Enter**), and a **stable accessible name** aligned with the [APG switch pattern](https://www.w3.org/WAI/ARIA/apg/patterns/switch/) (set **`aria-label`** or **`aria-labelledby`** on the host; name does not change with state; state is conveyed via `aria-checked`). Visible on/off copy is **not** used as the accessible name.
+**ix-toggle:** Exposes correct **switch** semantics (`role`, `aria-checked`, `aria-disabled`), keyboard interaction (**Space** / **Enter**), and a **stable accessible name** (set **`aria-label`** or **`aria-labelledby`** on the host; name does not change with state; state is conveyed via `aria-checked`). Visible on/off copy is **not** used as the accessible name.
--- a/.changeset/toggle-accessibility-improvements.md
+++ b/.changeset/toggle-accessibility-improvements.md
@@ -2,4 +2,4 @@
 '@siemens/ix': patch
 ---
 
-Improve **ix-toggle** accessibility (switch semantics, keyboard support, ARIA) and add automated accessibility coverage.
+**ix-toggle:** Exposes correct **switch** semantics for assistive technologies (`role`, `aria-checked`, `aria-disabled`), supports expected **keyboard** interaction, and applies **`aria-label` / `aria-labelledby` and related host ARIA** on the control so screen readers get an accurate name.
--- a/packages/core/src/components/toggle/toggle.tsx
+++ b/packages/core/src/components/toggle/toggle.tsx
@@ -180,7 +180,7 @@ export class Toggle implements IxFormComponent<string> {
       <Host
         role="switch"
         tabindex={this.disabled ? -1 : 0}
-        aria-label={this.hostElement.getAttribute('aria-label') ?? 'Toggle'}
+        aria-label={this.hostElement.getAttribute('aria-label')}
         aria-checked={a11yBoolean(this.checked)}
         aria-disabled={a11yBoolean(this.disabled)}
         class={{
--- a/packages/core/src/components/toggle/toggle.tsx
+++ b/packages/core/src/components/toggle/toggle.tsx
@@ -200,9 +200,8 @@ export class Toggle implements IxFormComponent<string> {
           }
         }}
       >
-        <label
+        <div
           class="wrapper"
-          role="presentation"
           onClick={() => {
             if (!this.disabled) {
               this.onCheckedChange(!this.checked);
@@ -215,6 +214,7 @@ export class Toggle implements IxFormComponent<string> {
               checked: this.checked,
               indeterminate: this.indeterminate,
             }}
+            aria-hidden="true"
           >
             <div class="slider"></div>
           </div>
@@ -234,7 +234,7 @@ export class Toggle implements IxFormComponent<string> {
           {!this.hideText && (
             <ix-typography class="label">{toggleText}</ix-typography>
           )}
-        </label>
+        </div>
       </Host>
     );
   }
--- a/packages/core/src/components/toggle/toggle.tsx
+++ b/packages/core/src/components/toggle/toggle.tsx
@@ -176,11 +176,15 @@ export class Toggle implements IxFormComponent<string> {
     if (this.indeterminate) {
       toggleText = this.textIndeterminate;
     }
+
+    const ariaLabelAttr = this.hostElement.getAttribute('aria-label');
+    const ariaLabel = ariaLabelAttr || toggleText;
+
     return (
       <Host
         role="switch"
         tabindex={this.disabled ? -1 : 0}
-        aria-label={this.hostElement.getAttribute('aria-label')}
+        aria-label={ariaLabel}
         aria-checked={a11yBoolean(this.checked)}
         aria-disabled={a11yBoolean(this.disabled)}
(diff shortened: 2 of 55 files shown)


packages/core/src/components/toggle/toggle.tsx
```diff
@@ -176,46 +176,65 @@ export class Toggle implements IxFormComponent<string> {
...
-              this.onCheckedChange((event.target as HTMLInputElement).checked)
-            }
+            onFocus={() => this.hostElement.focus()}
+            onClick={(e) => {
+              e.preventDefault();
+              e.stopPropagation();
+            }}
           />
           {!this.hideText && (
             <ix-typography class="label">{toggleText}</ix-typography>
```
lzeiml: This introduces redundant text to the a11y tree, making screenreader read "On  switch  on". We should hide it using aria-hidden.
see: https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch/

packages/core/src/components/toggle/toggle.tsx
```diff
@@ -176,46 +176,65 @@ export class Toggle implements IxFormComponent<string> {
...
+
+    const ariaLabelAttr = this.hostElement.getAttribute('aria-label');
+    const ariaLabel = ariaLabelAttr || (this.hideText ? toggleText : undefined);
+
     return (
       <Host
         role="switch"
+        tabindex={this.disabled ? -1 : 0}
+        aria-label={ariaLabel}
         aria-checked={a11yBoolean(this.checked)}
```
lzeiml: For indeterminate state the screenreader currently reads "off". aria-checked also supports the value "mixed" for this case.

lzeiml (dismissed):

packages/core/src/components/toggle/toggle.tsx
alexkaduk: thanks for highlighting. here https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-checkbox/ it states 
image

packages/core/src/components/toggle/toggle.tsx
alexkaduk: @lzeiml  based on this https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/switch_role#description
the switch role does NOT support "mixed" value

it can be done like
`const ariaChecked = this.indeterminate
      ? 'mixed'
      : a11yBoolean(this.checked);

    return (
      <Host
        role={this.indeterminate ? 'checkbox' : 'switch'}
        aria-checked={ariaChecked}
`
might it make sense? or no updates might be a way to go?

packages/core/src/components/toggle/toggle.tsx
alexkaduk: at the same time it is required to have aria-label. so did it like this `aria-hidden={ariaLabelAttr ? 'true' : undefined}`

packages/core/src/components/toggle/toggle.tsx
lzeiml: @alexkaduk I think the role should be constant. Considering indeterminate state is possible, checkbox role would be warranted here as a general role.

packages/core/src/components/toggle/toggle.tsx
lzeiml: @alexkaduk Will discuss this with a11y team and then we can make a decision

lzeiml: did some more research:
only aria-hidden=true if textOn, textOff and textIndeterminate are their default values. if one is not default, set aria-hidden=false. we could add 3 constants here for initializing the props and for the check.

packages/core/src/components/toggle/toggle.tsx
lzeiml: did some more research:
keep role=switch always. if the state is indeterminate aria-checked=false is fine. i checked other design systems they do it like that.

packages/core/src/components/toggle/toggle.tsx
alexkaduk: adjusted. @lzeiml pls check

.changeset/toggle-accessibility-improvements.md
```diff
@@ -0,0 +1,5 @@
+---
+'@siemens/ix': patch
+---
+
+Improve **ix-toggle** accessibility (switch semantics, keyboard support, ARIA) and add automated accessibility coverage.
```
danielleroux: > add automated accessibility coverage

Not relevant maybe give a more detailed description about the improvements

packages/core/src/components/toggle/toggle.tsx
```diff
@@ -64,17 +77,17 @@ export class Toggle implements IxFormComponent<string> {
...
    * Text for off state
    */
-  @Prop() textOff = 'Off';
+  @Prop() textOff = DEFAULT_TEXT_OFF;
 
   /**
    * Text for indeterminate state
    */
-  @Prop() textIndeterminate = 'Mixed';
+  @Prop() textIndeterminate = DEFAULT_TEXT_INDETERMINATE;
```
danielleroux: I would consider to keep it as a inline string because this will rendered in the documentation like:

```
<ApiTable id="property-textOn">
   <!-- others -->

  <ApiTable.Code name="Default">
    {`DEFAULT_TEXT_ON`}
  </ApiTable.Code>

</ApiTable>
```

Without having a feature to resolve the default value before rendering to markdown this makes not sense.

packages/core/src/components/toggle/toggle.tsx
```diff
@@ -176,46 +190,73 @@ export class Toggle implements IxFormComponent<string> {
...
+        aria-label={ariaLabel}
         aria-checked={a11yBoolean(this.checked)}
         aria-disabled={a11yBoolean(this.disabled)}
+        aria-required={a11yBoolean(this.required)}
         class={{
           disabled: this.disabled,
         }}
         onBlur={() => this.ixBlur.emit()}
         onFocus={() => (this.touched = true)}
+        onKeyDown={(e: KeyboardEvent) => {
```
danielleroux, line 217: Using inheritAriaAttributes remove elements from the host and applies it back again. Which cause an change event.

Complete idea of usage of inherit aria attributes makes no sense if you are having them on the host element.

1. `ix-toggle` aria-label
2. InheritAriaAttributesMixin => event for aria-label change on host. Store the value and removes it `removeAttribute('aria-label')`
3. You apply it again in the render function which cause again with point 2.

Maybe related but hard to say with the mixin applied

Image

aria-label getting out of sync

packages/core/src/components/toggle/toggle.tsx
```diff
@@ -176,46 +190,73 @@ export class Toggle implements IxFormComponent<string> {
...
             disabled={this.disabled}
             indeterminate={this.indeterminate}
             checked={this.checked}
-            tabindex={0}
-            aria-hidden={a11yBoolean(true)}
-            aria-checked={a11yBoolean(this.checked)}
-            onChange={(event) =>
-              this.onCheckedChange((event.target as HTMLInputElement).checked)
-            }
+            onFocus={() => this.hostElement.focus()}
```
danielleroux: You set tabindex to `-1` and add onFocus listener does not really does anything now because you disabled focusing. What was your idea behind this?

packages/core/src/components/toggle/toggle.tsx
```diff
@@ -176,46 +190,73 @@ export class Toggle implements IxFormComponent<string> {
...
               indeterminate: this.indeterminate,
             }}
-            onClick={() => this.onCheckedChange(!this.checked)}
-            tabindex={-1}
             aria-hidden="true"
           >
             <div class="slider"></div>
-          </button>
+          </div>
           <input
```
danielleroux: Regarding your changes I would also challenge to discuss why you still using `<input>` here. With the current refactorings its only style related anymore and this can be achieved with a with a `div.class`

packages/core/src/components/toggle/toggle.tsx
```diff
@@ -176,46 +190,73 @@ export class Toggle implements IxFormComponent<string> {
...
-              this.onCheckedChange((event.target as HTMLInputElement).checked)
-            }
+            onFocus={() => this.hostElement.focus()}
+            onClick={(e) => e.preventDefault()}
           />
           {!this.hideText && (
-            <ix-typography class="label">{toggleText}</ix-typography>
+            <ix-typography
+              class="label"
+              aria-hidden={isDefaultLabels ? 'true' : undefined}
```
danielleroux:
```suggestion
              aria-hidden={a11yBoolean(useToggleTextAsLabel)}
```

Utility function resolves string "true" or "false"

testing/framework-tests/src/main.ts
danielleroux, line 1: > testing/framework-tests/src/main.ts adjusted as needed.

What is needed here? Please explain in more detail

danielleroux (changes requested):

.changeset/toggle-accessibility-improvements.md
alexkaduk: adjusted. thanks

packages/core/src/components/toggle/toggle.tsx
alexkaduk: updated. thanks

packages/core/src/components/toggle/toggle.tsx
alexkaduk: removed hidden input

packages/core/src/components/toggle/toggle.tsx
alexkaduk: removed hidden input

testing/framework-tests/src/main.ts
alexkaduk: updated, thanks

packages/core/src/components/toggle/toggle.tsx
alexkaduk: now always uses aria-hidden="true" on the typography so the visible on/off copy is never the name

https://www.w3.org/WAI/ARIA/apg/patterns/switch/
> Important: it is critical the label on a switch does not change when its state changes.

packages/core/src/components/toggle/toggle.tsx
alexkaduk: updated, the switch now keeps a stable host name (aria-label / aria-labelledby); state is only in aria-checked

danielleroux (approved):

PR #2409, opened 2026-02-24
Updated Readonly date and timeinput 

varun-srinivasa:

When working with read-only forms, I noticed something unexpected happens during validation. If there is an error in any input value, the expected behavior is that the input remains non-interactive however when the error appears the input becomes selectable/interactive. only in date and time input

GitHub Issue Number: [IX-3909](https://agileworld.siemens.cloud/jira/browse/IX-3909)

Now even in readonly form the date and time input components are not interactive

## 🏁 Checklist


packages/core/src/components/date-input/date-input.tsx
```diff
@@ -576,7 +582,11 @@ export class DateInput
         </ix-field-wrapper>
         <ix-dropdown
           data-testid="date-dropdown"
-          trigger={this.inputElementRef.waitForCurrent()}
+          trigger={
+            this.isInteractive
+              ? this.inputElementRef.waitForCurrent()
+              : undefined
+          }
```
danielleroux: 1. `isInteractive` is no state variable
2. Overall better would be to prevent the showchanged or showchanging event

(This counts for all changes in this pr)

packages/core/src/components/datetime-input/datetime-input.tsx
```diff
@@ -620,7 +624,7 @@ export class DatetimeInput
               event.preventDefault();
             }
           }}
-          onFocus={() => {
+          onFocus={async () => {
```
danielleroux: Does not need to be async

danielleroux (changes requested):
Changeset missing

packages/core/src/components/datetime-input/datetime-input.tsx
varun-srinivasa: Removed the unnecessary async keyword from all three onfocus handlers since none of them use await.

packages/core/src/components/date-input/date-input.tsx
```diff
@@ -583,6 +589,11 @@ export class DateInput
           enableTopLayer={this.enableTopLayer}
           suppressOverflowBehavior
           show={this.show}
+          onShowChange={(event) => {
+            if (!this.isInteractive) {
+              event.preventDefault();
+            }
+          }}
```
dmytro-halimov, line 596: if the component becomes `readonly` or `disabled` while the dropdown is already opened, closing will also be prevented. We could change this to:

```
if (event.detail && !this.isInteractive) {
  event.preventDefault();
}
```

This would match the existing pattern in ix-select, where only opening is blocked for readonly/disabled states.

packages/core/src/components/datetime-input/datetime-input.tsx
```diff
@@ -806,6 +810,11 @@ export class DatetimeInput
           show={this.show}
           suppressOverflowBehavior
           trigger={this.inputElementRef.waitForCurrent()}
+          onShowChange={(event) => {
+            if (!this.isInteractive) {
+              event.preventDefault();
+            }
+          }}
```
dmytro-halimov, line 817: The same `isInteractive` getter and `onShowChange` guard are now repeated more than few times across components. Since this is shared date/time picker behavior, could we extract this behavior into picker-input.util.ts?

dmytro-halimov (changes requested):
Could you please also add a changeset file describing all these changes?

PR #2415, opened 2026-03-01
feat(pill): improve accessibility #IX-3654

alexkaduk:

`ix-pill` did not surface host ARIA on the visible pill surface inside shadow DOM, so names and roles set on `<ix-pill>` were easy to miss in the accessibility tree. Leading icons were not consistently treated as decorative when the pill already had text or a host label.

## What is the new behavior?

- **Consumers:** `aria-label`, `aria-labelledby`, and other host ARIA handled by the pill are applied to the inner pill container, so assistive technologies see the intended name and semantics. If the pill has `aria-label` or `aria-labelledby` and no explicit host `role`, the container gets **`role="group"`** by default; an explicit host **`role`** still wins.
- **Icons:** Without **`aria-label-icon`**, the leading **`ix-icon`** is **`aria-hidden`** (decorative). With **`aria-label-icon`**, the icon gets an accessible name.
- **Examples:** Preview examples (HTML, **Angular module + standalone**, React, Vue) and pill visual fixtures use accessible patterns where needed (e.g. icon-only pill with host **`aria-label`**); custom styling uses design tokens where applicable.

### Implementation & verification (for reviewers)

- **Core:** `packages/core/src/components/pill/pill.tsx` — `InheritAriaAttributesMixin`, container spread + `role` logic, `a11yBoolean` for icon `aria-hidden`.
- **Tests:** `packages/core/src/components/pill/test/pill.ct.ts` (host → container ARIA, decorative vs labeled icon, combined attributes).
- **Storybook:** `packages/storybook-docs/src/stories/pill/accessibility/a11y-pill.stories.ts`.
- **Framework CI:** `pill` / `pill-variants` in `testing/framework-tests/tests/working.spec.ts` and `working-with-axe.spec.ts`, with updated ARIA snapshots and generated Playwright specs.

## 🏁 Checklist

- [ ] 🗺️ Internationalization (i18n) - no hard coded strings *(confirm: no new user-visible copy beyond examples)*
- [ ] 📲 Responsiveness - components handle viewport changes and content overflow gracefully *(confirm unchanged / still OK)*
- [ ] 🧪 Unit tests were added/updated and pass (`pnpm test`) *(includes `pill.ct.ts` / core tests)*


.changeset/pill-accessibility-improvements.md
```diff
@@ -0,0 +1,5 @@
+---
+'@siemens/ix': patch
+---
+
+Improve **ix-pill** accessibility with ARIA attribute inheritance on the pill surface and additional automated accessibility tests.
```
danielleroux: Please describe the change for the changelog. Internal architecture and testing like inheritance and tests is not necessary to mention in a changeset

packages/angular-standalone-test-app/src/preview-examples/pill-variants.ts
danielleroux, line 1: I do not like the idea of simplify the example. Its an example for developer to copy code fast. If you do this in a more complex way like with a "@for" the user needs additional adaptions.

(Includes all examples not only this angular example)

packages/angular-standalone-test-app/src/preview-examples/pill-variants.ts
```diff
@@ -21,87 +21,52 @@ import {
...
+            >{{ row.label }}</ix-pill>
+          </ix-col>
+        </ix-row>
+      }
       <ix-row>
         <ix-col>
           <ix-pill
             variant="custom"
-            color="color-soft-text"
+            pill-color="white"
```
danielleroux: I know we did this with purple, but I would change this color value to a token which we provide. This makes a good guidance regarding design tokens

(Includes all examples, just mention it here once)

packages/core/src/components/pill/pill.tsx
```diff
@@ -153,6 +176,7 @@ export class Pill implements IxComponentInterface {
               name={this.icon}
               size={'16'}
               aria-label={this.ariaLabelIcon}
+              aria-hidden={this.ariaLabelIcon ? undefined : 'true'}
```
danielleroux:
```suggestion
              aria-hidden={a11yBoolean(this.ariaLabelIcon)}
```

danielleroux (changes requested):

packages/angular-standalone-test-app/src/preview-examples/pill-variants.ts
alexkaduk: reverted. used simple markup

packages/core/src/components/pill/pill.tsx
alexkaduk: adjusted. thanks

.changeset/pill-accessibility-improvements.md
alexkaduk: updated, thanks

packages/core/src/components/pill/pill.tsx
```diff
@@ -126,6 +150,8 @@ export class Pill implements IxComponentInterface {
         }}
       >
         <div
+          {...this.inheritAriaAttributes}
```
danielleroux: Why is it necessary to have the aria attributes inherit from the Host to the inner div? Typically you use this pattern if you have a focusable element inside the host element to inherit the host attributes to to the focusable element.

For the ix-pill this makes no sense because its a none focusable element. Which mean i would suggest to keep the existing host attributes simply on the host element

danielleroux (changes requested):

packages/core/src/components/pill/pill.tsx
alexkaduk: removed InheritAriaAttributesMixin from ix-pill

packages/storybook-docs/src/stories/pill/accessibility/a11y-pill.stories.ts
```diff
@@ -0,0 +1,212 @@
...
+    return html`
+      <div>
+        <ix-pill
+          aria-label="System status"
+          aria-describedby="status-desc"
+          role="status"
+          variant="success"
+          >Online</ix-pill
+        >
+        <span id="status-desc" hidden>
```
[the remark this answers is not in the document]
alexkaduk: hidden was replaced for valid aria-describedby behavior

packages/storybook-docs/src/stories/pill/accessibility/a11y-pill.stories.ts
```diff
@@ -175,7 +175,10 @@ export const WithDescription: Story = {
           variant="success"
           >Online</ix-pill
         >
-        <span id="status-desc" hidden>
+        <span
+          id="status-desc"
+          style="position: absolute; width: 1px; height: 1px; padding: 0; margin: -1px; overflow: hidden; clip: rect(0, 0, 0, 0); white-space: nowrap; border: 0;"
+        >
           System has been operational for 48 hours without interruption.
         </span>
```
danielleroux: Why did you try to hide this? Just leave it as element in the view

danielleroux (changes requested):

packages/storybook-docs/src/stories/pill/accessibility/a11y-pill.stories.ts
alexkaduk: removed hiding. it was added based on Copilot’s suggestion

danielleroux (approved):

Issue #798, opened 2023-10-03
 Toast should be displayed above modal


viktorhajer:
### What happened?

Currently the toast is behind the modal window -> so it is not possible to display toast from dialog.

### What type of frontend frameware are you seeing the problem on?

Angular

### Which version of iX do you use?

v2.0.3

### Code to produce this issue.

```shell
export class TestComponent {

  constructor(private readonly modalService: ModalService,
              private readonly toastService: ToastService,) {
    this.toastService.setPosition('top-right');
  }

  open(){
    this.toastService.show({title: 'title', message: 'msg'});
    this.modalService.open({content: DialogComponent, size: '720'});
  }
}
```

viktorhajer:
This is still an issue. I updated the version in the description.

viktorhajer:
This is still an issue in 2.0.4. Any update?

silviowolf:
No update 2.0.4, is in backlog but not planned yet.

viktorhajer:
Okay, thank you.

viktorhajer:
Hi @silviowolf, as I see it is planned now. Can we expect it in the next release (2.1.0)?

supreethaPavank:
+1 . I am also facing this issue

matthiashader:
[IX-609]

raphaelgodart:
What's the status on resolving this problem? This bug has been open for quite a while and a fix is become rather urgent. Our team has an upcoming software release. Currently, notifications to the user (including errors) are currently invisible for the user when a toast is triggered from a modal.

flxlst09:
Hello @raphaelgodart  We understand the importance of this functionality and apologize for the delay in our response. To provide some transparency regarding the challenges we are facing with this issue, here are some of the key considerations we have been working through:

- Before displaying a toast message, we would need to check if a modal is open. If a modal is open, ensure the toast message is displayed above the modal layer.
- If a toast message is already open and a modal is triggered, we would need to dynamically adjust the toast’s layer to ensure it remains visible as we want to avoid having toast messages open on multiple layers.
- We should also define clear rules for user interactions with toast messages while a modal is open. For example, if an action in a toast message requires closing the modal.
- Given these complexities and the urgency of your project, we have re-prioritized this issue into our Q1 2025 planning. We are committed to finding a robust solution and will keep you updated on our progress.

As a temporary workaround, you could consider not showing the backdrop when opening modal dialogs. This would not entirely solve the problem, but toast messages would be visible.

[...]

PR #2471, opened 2026-03-31
feat(modal): add non-blocking dialog mode IX-3843

alexkaduk:

`ix-modal` always behaved as a blocking modal: `showModal()` on the native `<dialog>`, lightbox backdrop, focus trap, and `aria-modal="true"`. There was no supported way to open a non-modal panel where the rest of the page stays interactive and focus can move outside the dialog.

GitHub Issue Number: #798

- Adds **`isNonBlocking`** on `ix-modal` and **`ModalConfig.isNonBlocking`** for `showModal()`: uses `dialog.show()` instead of `showModal()`, no blocking lightbox/focus trap, and **`aria-modal="false"`** on the inner `<dialog>`.
- **Initial focus** for non-blocking modals is handled explicitly after open (autofocus / header close / dialog host), including timing after slide-in where needed.
- **Utilities** in `utils/modal/modal.ts` (e.g. `applyIxModalNonBlockingInitialFocus`, blocking-path autofocus helpers) keep behavior consistent between blocking and non-blocking flows.

## 🏁 Checklist


packages/core/src/components/modal/test/modal.ct.ts
```diff
@@ -18,11 +18,11 @@ import {
 } from '@siemens/ix-icons/icons';
 
 declare global {
-  interface Window {
-    dismissModal: typeof dismissModal;
-    showModal: typeof showModal;
-    __counter: number;
-  }
+  var showModal: typeof import('./../../utils/modal').showModal;
+  var dismissModal: typeof import('./../../utils/modal').dismissModal;
```
danielleroux: Is it really necessary to to inline imports here?

packages/core/src/components/modal/test/modal.ct.ts
```diff
@@ -46,7 +46,19 @@ async function setupModalEnvironment(page: Page) {
...
+      typeof globalThis.dismissModal === 'function' &&
+      globalThis.showMessage != null,
+    { timeout: 15_000 }
+  );
+}
+
+async function waitForModalDialogOpen(page: Page) {
+  const dialog = page.locator('ix-modal dialog');
+  await expect(dialog).toBeVisible({ timeout: 5000 });
+  await expect(dialog).toHaveCSS('opacity', '1');
```
danielleroux: Do we need here the opacity check? Would be prefer to have a some kind of aria attribute here or maybe a class indicator.

packages/core/src/components/modal/test/modal.ct.ts
```diff
@@ -279,55 +269,337 @@ regressionTest.describe('closeOnBackdropClick = true', () => {
...
+        globalThis.showModal({
+          content: elm,
+          isNonBlocking: true,
+          animation: false,
+        });
+      });
+
+      await waitForModalDialogOpen(page);
+      await expect(page.locator('ix-modal')).toHaveClass(/non-blocking/);
+      await expect(page.locator('ix-modal dialog')).toHaveAttribute(
```
danielleroux: Just a small suggestion maybe not worth to change but instead of a hard css selector would it make sense to use here the getByRole?

packages/core/src/components/modal/test/modal.ct.ts
```diff
@@ -279,55 +269,337 @@ regressionTest.describe('closeOnBackdropClick = true', () => {
...
+          content: elm,
+          isNonBlocking: true,
+          animation: false,
+        });
+      });
+
+      await waitForModalDialogOpen(page);
+      await expect(page.locator('ix-modal ix-button[autofocus]')).toBeFocused({
+        timeout: 5000,
+      });
```
danielleroux: Use here an accessibility related selector e.g getByRole('button', { hasText: 'OK'})

packages/core/src/components/modal/test/modal.ct.ts
```diff
@@ -279,55 +269,337 @@ regressionTest.describe('closeOnBackdropClick = true', () => {
...
+
+      await waitForModalDialogOpen(page);
+      await page.locator('ix-modal ix-icon-button.modal-close').focus();
+      await page.keyboard.press('Escape');
+      await expect(page.locator('ix-modal dialog')).not.toBeVisible();
+    }
+  );
+
+  regressionTest(
+    'non-blocking modal should not close on Escape when disableEscapeClose is true',
```
danielleroux: Regarding the w3c `disableEscapeClose` does prevent to implement good accessibility related modals/dialogs. Lets remove this property completely (major change then not minor + breaking changes entry), to be always 100% accessibility conform regarding closing behaviour

packages/core/src/components/modal/modal.tsx
```diff
@@ -21,9 +21,17 @@ import { animate } from 'animejs';
...
 import { OnListener } from '../utils/listener';
+import { tryFocusElement } from '../utils/focus/focus-utilities';
+import { IX_MODAL_AUTOFOCUS_SELECTOR } from '../utils/modal/modal';
 import { waitForElement } from '../utils/waitForElement';
 import { IxModalSize } from './modal.types';
 
+/** One microtask so slotted / framework children exist
+ * before we request an animation frame and query `[autofocus]`. */
+function deferToNextMicrotask(): Promise<void> {
+  return new Promise((resolve) => queueMicrotask(resolve));
```
danielleroux: I would stick to `requestAnimationFrame`, cause stencil call the **initial** rendering inside a microtask (https://github.com/stenciljs/core/blob/76376d5acdddfccb80216aba69deb36c1e2c3704/src/runtime/update-component.ts#L42) but the later updates are inside a animationFrame (https://github.com/stenciljs/core/blob/76376d5acdddfccb80216aba69deb36c1e2c3704/src/client/client-task-queue.ts#L23)

Also frameworks like angular (at least older versions) makes the changedetection based on the animationFrame.
For React (vdom) tbh not quit sure how this works but using animationFrame should also be fine.

Any additional information from your side why `queueMicrotask` is choosen?

packages/core/src/components/utils/focus/focus-utilities.ts
```diff
@@ -63,6 +63,26 @@ export function queryElements(
   return items;
 }
 
+export type DomFocusOptions = Parameters<HTMLElement['focus']>[0];
+
+/** `focusVisible` is supported in Firefox 104+; omitted from some TypeScript DOM typings. */
+export type IxFocusOptions = DomFocusOptions & { focusVisible?: boolean };
+
+export function tryFocusElement(
+  element: Element | HTMLElement | null | undefined,
```
danielleroux:
```suggestion
  element: Element | HTMLElement
```

> tryFocusElement(document.getElementById('not existing id'));

I would suggest not allow this and make it more strict to handle undefined/null elements further up in the call tree

packages/core/src/components/utils/focus/focus-utilities.ts
```diff
@@ -63,6 +63,26 @@ export function queryElements(
...
+  options?: IxFocusOptions
+): boolean {
```
danielleroux: Whats the purpose behind the `boolean` return? The return value is never used in the current code base.

packages/core/src/components/modal/modal.tsx
```diff
@@ -173,13 +188,24 @@ export class Modal {
...
     }
   }
 
+  private async scheduleInitialAutofocus() {
+    await deferToNextMicrotask();
+    requestAnimationFrame(() => {
+      const direct = this.hostElement.querySelector<HTMLElement>(
+        IX_MODAL_AUTOFOCUS_SELECTOR
+      );
+      tryFocusElement(direct, { focusVisible: true });
```
danielleroux: Did you checked if the new introduced focusVisible option has any side effect regarding the mixin here: https://github.com/siemens/ix/blob/main/packages/core/src/components/utils/internal/mixins/setup.mixin.ts

packages/core/src/components/modal/test/modal.ct.ts
danielleroux, line 1: Maybe a testcase missing regarding autofocus of the modal-header e.g

```html
<ix-modal>
   <ix-modal-header autofocus>
```

In this case we can autofocus the close button which is otherwise not possible. To get this working we need to delegate the modal-header focus:
```
  shadow: {
    delegatesFocus: true,
  }
```

danielleroux (changes requested):

packages/core/src/components/modal/test/modal.ct.ts
```diff
@@ -18,11 +18,11 @@ import {
...
-    dismissModal: typeof dismissModal;
-    showModal: typeof showModal;
-    __counter: number;
-  }
+  var showModal: typeof import('./../../utils/modal').showModal;
+  var dismissModal: typeof import('./../../utils/modal').dismissModal;
+  var showMessage: any;
+  var __counter: number;
+  var __nbBgClick: boolean | undefined;
 }
```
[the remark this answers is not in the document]
alexkaduk: updated to call to globalThis.showModal(...)

packages/core/src/components/modal/test/modal.ct.ts
alexkaduk: replaced the opacity assertion with a `modal-open-settled` class on <dialog>

set when the open animation completes (and immediately when animation duration is 0)

waitForModalDialogOpen now asserts that class so we still wait for a settled panel without depending on CSS opacity

packages/core/src/components/modal/test/modal.ct.ts
alexkaduk: replaced ix-modal dialog with page.getByRole('dialog') through a modalPanel(page) helper

packages/core/src/components/modal/test/modal.ct.ts
alexkaduk: updated to getByRole('button', { name: 'OK' }) for the focus assertion. applied the same pattern in “button receives focus on load”

packages/core/src/components/modal/test/modal.ct.ts
alexkaduk: removed disableEscapeClose.

escape / <dialog> cancel always call dismissModal(); blocking is done with beforeDismiss.

showModalLoading now sets beforeDismiss: () => false; finish() still uses closeModal. 

documented in BREAKING_CHANGES/v5.md and the changeset (major bump).

packages/core/src/components/modal/modal.tsx
alexkaduk: Agreed — dropped queueMicrotask and now use double requestAnimationFrame before querying [autofocus], so timing lines up with Stencil’s rAF-based updates instead of a one-off microtask.

On double rAF: React needs two frames (e.g. commit then paint) before slotted children exist.

packages/core/src/components/modal/modal.tsx
alexkaduk: without { focusVisible: true }, script-driven focus when opening the modal often doesn’t get the same :focus-visible treatment as Tab/keyboard focus. on my machine I only reliably reproduce that in Firefox (no outline without the option; with focusVisible: true it behaves as expected).

native focus ring and IX focus-visible classes can both apply, but they come from different mechanisms (browser vs keyboard-mode styling), so they shouldn’t normally conflict - I could not find a bad double case locally. would be good if someone else can sanity-check in Chrome/Safari/FF on their side

packages/core/src/components/utils/focus/focus-utilities.ts
alexkaduk: updated. thanks

packages/core/src/components/utils/focus/focus-utilities.ts
alexkaduk: dropped the boolean return

packages/core/src/components/modal/test/modal.ct.ts
alexkaduk: updated ix-modal-header and added test

danielleroux (approved):

Issue #2053, opened 2025-08-07
[Feature Request] ix-upload: Support for Folder upload


itushardGIt:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### Suggestion / feature request

Current ix-upload component support only File selection to upload. We have requirement from Siemens Client (DI FA) in CDS area that User should able to select folder (entire hierarchy inside selected folder) and file to upload.

As this is client requirement, can we take this as Feature Enhancement.

Below is reference screenshot:
<img width="787" height="482" alt="Image" src="https://github.com/user-attachments/assets/d02d6623-429f-497c-aeae-54a95e3f660d" />

Feel free to reach me for more details.

kathrinschalber:
bump - already worked on (see linked PR above)

kathrinschalber:
Was already resolved in V5.0 in the linked PR

PR #2478, opened 2026-04-03
feat(core/upload): add folder upload

khathija-ahamadi:
The ix-upload component only supports selecting individual files via the native file dialog or drag-and-drop. Entire folder for upload is not supported.

GitHub Issue Number: #2053 and #2354

- A new 'webkitdirectory' prop lets users select or drag-and-drop an entire folder instead of individual files.
- When enabled, the file dialog opens in folder-selection mode and all files inside the chosen folder are included.
- When a folder is selected via the 'Upload folder...'  button, the browser shows a built-in security confirmation dialog asking the user to confirm access to the folder's contents. This warning is not shown when the folder is dragged and dropped onto the upload area. This behavior remains like that of the native HTML.
- A component test verifies the folder upload mode works correctly.

## 🏁 Checklist


packages/core/src/components/upload/upload.tsx
```diff
@@ -39,6 +39,11 @@ export class Upload {
    */
   @Prop() multiple = false;
 
+  /**
+   * If webkitdirectory is true the user can drop or select a folder containing one or more files
+   */
+  @Prop() webkitdirectory = false;
```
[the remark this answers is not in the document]
khathija-ahamadi: The button label is already customizable via the existing i18nUploadFile prop (attribute: i18n-upload-file). When using webkitdirectory, consumers can set i18n-upload-file="Upload folder…" or any localized string. No additional i18n property is needed since the component doesn't determine the label — the consumer does based on their context.

packages/core/src/components/upload/upload.tsx
```diff
@@ -239,6 +244,9 @@ export class Upload {
 
   render() {
     const disabled = this.disabled || this.state === UploadFileState.LOADING;
+    const directoryAttributes = this.webkitdirectory
+      ? ({ webkitdirectory: true, directory: true } as Record<string, boolean>)
+      : {};
```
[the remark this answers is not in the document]
khathija-ahamadi: Updated.

packages/core/src/components/upload/upload.tsx
```diff
@@ -249,6 +257,7 @@ export class Upload {
               this.state !== UploadFileState.LOADING && this.isFileOver,
             checking: this.state === UploadFileState.LOADING,
             disabled: this.disabled,
+            'folder-upload': this.webkitdirectory,
```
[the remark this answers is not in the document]
khathija-ahamadi: Updated.

packages/core/src/components/upload/upload.scss
```diff
@@ -73,6 +73,7 @@
 
     ix-button {
       margin-inline-start: vars.$default-space;
+      flex-shrink: 0;
```
nuke-ellington: The button's parent will not be flex layout container. What is the purpose of this change?

packages/core/src/components/upload/upload.tsx
nuke-ellington: The property name `webkitdirectory` is also an HTML attribute. We try to stay away from those to prevent conflicts/confusion. I would also prefer something more clear like [directoryUpload](https://element.siemens.io/components/forms-inputs/file-uploader/#SiFileUploaderComponent.directoryUpload).

nuke-ellington (dismissed):

packages/core/src/components/upload/upload.scss
khathija-ahamadi: Agreed. Reverted the change. It is working as expected.

packages/core/src/components/upload/upload.tsx
khathija-ahamadi: Suggested change updated.

packages/core/src/components/upload/upload.tsx
```diff
@@ -239,7 +244,16 @@ export class Upload {
...
   render() {
     const disabled = this.disabled || this.state === UploadFileState.LOADING;
-    const { 'aria-label': ariaLabel = 'Upload files', ...a11y } = this.a11y;
+    const directoryAttributes = this.directoryUpload
+      ? ({ webkitdirectory: true, directory: true, multiple: true } as Record<
+          string,
+          boolean
+        >)
+      : {};
+    const defaultAriaLabel = this.directoryUpload
```
danielleroux, line 263: directoryUpload changes the input behavior and even switches the default aria-label to "Upload folder", but the visible default copy still says "+ Drag files here or…" and "Upload file…". That means enabling folder mode still presents file-oriented UX unless every consumer manually overrides the texts. The new test also hides this by explicitly setting i18n-upload-file="Upload folder…", so the default mismatch is not covered.

danielleroux (changes requested):

packages/core/src/components/upload/upload.tsx
khathija-ahamadi: Thank you. I previously expected users to override the default texts, but now i have updated the component to ensure that when folder mode is enabled, both the labels and on-screen text does the correct action. Tests have been updated as well.

packages/core/src/components/upload/test/upload.ct.ts
```diff
@@ -16,3 +16,62 @@ regressionTest('renders', async ({ mount, page }) => {
...
+        hasWebkitDirectory: input.hasAttribute('webkitdirectory'),
+        hasDirectory: input.hasAttribute('directory'),
+      };
+    });
+
+    expect(attrs).not.toBeNull();
+    expect(attrs?.selectFileText).toBe('+ Drag folder here or…');
+    expect(attrs?.buttonText).toBe('Upload folder…');
+    expect(attrs?.hasWebkitDirectory).toBe(true);
+    expect(attrs?.hasDirectory).toBe(true);
```
danielleroux: Dont use eval. You can use the playwright native expect functions

packages/core/src/components/upload/test/upload.ct.ts
```diff
@@ -16,3 +16,62 @@ regressionTest('renders', async ({ mount, page }) => {
...
+          ?.querySelector('ix-button')
+          ?.textContent?.trim(),
+        selectFileText: element.shadowRoot
+          ?.querySelector('.upload-text')
+          ?.textContent?.trim(),
+      };
+    });
+
+    expect(texts?.selectFileText).toBe('Custom drag text');
+    expect(texts?.buttonText).toBe('Custom button');
```
danielleroux: Dont use eval. You can use the playwright native expect functions

packages/core/src/components/upload/upload.tsx
```diff
@@ -39,6 +39,11 @@ export class Upload {
    */
   @Prop() multiple = false;
 
+  /**
+   * If directoryUpload is true the user can drop or select a folder containing one or more files
+   */
+  @Prop() directoryUpload = false;
```
danielleroux, line 47: `@since` tag is missing

danielleroux (changes requested):

danielleroux (approved):

PR #2501, opened 2026-04-16
feat(core/dropdown): Rename 'enable-top-layer' prop to suppress-top-layer and set default value to false

dmytro-halimov:

Dropdown component uses 'enable-top-layer' prop.

GitHub Issue Number: #4133

- Prop renamed to 'suppress-top-layer' and default value set to false.
- Introduced mixin to control the renamed prop.

Snapshots are updated because of the new dropdown rendering default (native dialog element). So components that use dropdown were affected and therefore snapshots had to be updated.

## 🏁 Checklist


packages/core/src/components/menu-avatar-item/test/menu-avatar-item.ct.ts
```diff
@@ -34,22 +34,27 @@ regressionTest('Nested dropdowns', async ({ mount, page }) => {
...
 
-  await expect(dropdown1).toBeVisible();
+  await expect(dropdown1).toHaveClass(/hydrated/);
+  await expect(dropdownPanel(dropdown1)).toBeVisible();
 
   const dropdown2Trigger = dropdown1
     .locator('ix-dropdown-item')
     .filter({ hasText: 'SubMenuItem 4' });
-  await dropdown2Trigger.click();
+  await dropdown2Trigger.click({ force: true });
```
danielleroux: Why do we need to perform a forced click? Normally this indicates that something block a regular click.

packages/core/src/components/utils/internal/mixins/top-layer.mixin.ts
```diff
@@ -0,0 +1,35 @@
...
+
+export const TopLayerMixin = <B extends MixedInCtor<StencilLifecycle>>(
+  Base: B
+) => {
+  class TopLayerMixinCtor extends Base implements TopLayerMixinContract {
+    /**
+     * Suppress Popover API rendering for dropdown.
+     *
+     * @default false
+     * @since 4.3.0
```
danielleroux: Needs to be 5.0.0 not 4.3.0

packages/core/src/components/utils/internal/mixins/top-layer.mixin.ts
```diff
@@ -0,0 +1,35 @@
...
+     */
+    @Prop() suppressTopLayer: boolean = false;
+
+    get useTopLayer() {
+      return !this.suppressTopLayer;
+    }
```
danielleroux, line 31: Can you explain me why the getter here required?

danielleroux (changes requested):
- Changeset is missing regarding the components, also fixes (like select change) have to be documented
- Breakingchanges v5.md missing a migration guide (before/after) see existing ones
-

packages/core/src/components/utils/internal/mixins/top-layer.mixin.ts
dmytro-halimov: It's mainly a read-only alias for better readability and usage. !this.suppressTopLayer were used multiple times, so I've introduced this getter to avoid that scattered negations.
I have no strong opinion about it and ok to remove it if that's something you'd rather not have.

dmytro-halimov:
> * also fixes (like select change) have to be documented

@danielleroux Could you maybe provide some example on how to properly document the fix?

packages/core/src/components/utils/internal/mixins/top-layer.mixin.ts
dmytro-halimov: done

packages/core/src/components/menu-avatar-item/test/menu-avatar-item.ct.ts
dmytro-halimov: good catch, I've refactored it a bit to get rid of it

danielleroux:
> > * also fixes (like select change) have to be documented
> 
> @danielleroux Could you maybe provide some example on how to properly document the fix?

Create a changeset inside this folder: https://github.com/siemens/ix/tree/main/.changeset

packages/core/src/components/dropdown/dropdown.scss
nuke-ellington, line 1: Is this change really neccessary?

packages/core/src/components/select/test/select-controller.ts
```diff
@@ -9,6 +9,10 @@
 
 import { Locator, expect } from '@playwright/test';
 
+function dropdownPanel(dropdown: Locator): Locator {
+  return dropdown.locator('dialog').first();
+}
+
```
nuke-ellington: Import test util

packages/core/src/tests/utils/test/index.ts
```diff
@@ -44,3 +44,11 @@ export const getFormValue = async (
     return formData.get(key);
   }, key);
 };
+
+/**
+ * Popover surface for `ix-dropdown` in top-layer mode (`<dialog popover>`).
+ * Uses `.first()` so nested `ix-dropdown` dialogs inside the menu do not break strict locators.
+ */
+export function dropdownPanel(dropdown: Locator): Locator {
```
nuke-ellington: IMO the name is not very clear. Consider something like getDialog?

testing/visual-testing/tests/dropdown/dropdown.e2e.ts
```diff
@@ -74,15 +79,28 @@ regressionTest.describe('dropdown', () => {
...
-    await page.waitForSelector('.dropdown-menu.show');
+    await waitForOpenDropdownPanel(page);
 
     await expect(page).toHaveScreenshot();
   });
 
   regressionTest('centered overflow', async ({ page }) => {
     await page.goto('dropdown/centered-overflow');
-    const lastItem = await page.locator('.dropdown-item').last();
-    await lastItem.scrollIntoViewIfNeeded();
```
nuke-ellington, line 85: Is this change really neccessary?

nuke-ellington (changes requested):
Generally we try to avoid too many comments explaining logic and changes. Please remove them unless absolutely neccessary.

packages/core/src/components/dropdown/dropdown.scss
dmytro-halimov: yes, without it we have a visual bug (host expands with applied padding) caused by the styles applied to host

testing/visual-testing/tests/dropdown/dropdown.e2e.ts
dmytro-halimov: without this change with the new underlying structure this test was failing, had to update it to make it work

dmytro-halimov:
These changes to the dropdown component will be released within the next major version

PR #2532, opened 2026-05-05
fix(core/date-picker): Set correct roles for date-picker grid and time-picker listbox

lzeiml:
IX-4254

When NVDA is running the keyboard navigation for date-picker/time-picker does not work as expected (visual focus does not move and order is incorrect)

date-picker: Set the correct aria-roles (grid, gridcell), for the days container and the days elements themselves.
time-picker: Set role listbox for the columns

## 🏁 Checklist


danielleroux (changes requested):
# Datepicker (VoiceOver) overcommunicate

Datepicker still announce to much:

https://github.com/user-attachments/assets/9d28fa9c-d773-4034-876e-b36a846cc2da


# Timepicker does not tell which hour is selected

Open timpicker oder timeinput. Tab to Hour use the keyboard and navigate does not tell if the focused item is the current selected one

# Hour container is selectable

Currently the container holding the hours can be focused and scrolled via keyboard

Image

I would agree if every container is focusable to scroll without changing the selection (aria label is then missing i think), but currently this is only working with hour

danielleroux (approved):

PR #2541, opened 2026-05-07
fix(vue,react,angular-standalone): improve treeshaking of the library 

danielleroux:

- Wrap utility function with depnendency wrapper to define custom elements if needed
- Currently not working natively for dist-custom-element output without a framework wrapper
- Reduce internal components to minimum required ones  

## 🏁 Checklist


packages/react-test-app/src/preview-examples/loading.tsx
```diff
@@ -11,9 +11,11 @@ import { ModalLoadingContext } from '@siemens/ix';
 import { IxButton, showModalLoading } from '@siemens/ix-react';
 
 export default () => {
-  const startLoading = () => {
+  const startLoading = async () => {
```
dmytro-halimov, line 14: i've noticed that for html-test-app preview example is not updated – it still uses the synchronous `showModalDialog`

packages/core/src/components/utils/dependency-function.ts
```diff
@@ -0,0 +1,112 @@
...
+    };
+
+    const dependencyFunction = (async (...args: Parameters<TCallback>) => {
+      await ensureDependencies();
+      return callback(...args);
+    }) as DependencyFunction<TCallback, TDependencies>;
+
+    Object.defineProperties(dependencyFunction, {
+      dependencies: {
+        enumerable: true,
```
dmytro-halimov: not sure if this is inentional, but seems like it also should be non-enumerable (also given the fact that `dependencyDefinitions` are frozen)

packages/core/scripts/build/vue-output-target.ts
```diff
@@ -0,0 +1,112 @@
...
+    return '';
+  }
+
+  return `, [
+  ${values.map((value) => `'${value}'`).join(',\n  ')}
+]`;
+}
+
+function generateVueComponent(
+  component: any,
```
dmytro-halimov: we can utilize more specific types that Stencil ships for `component` here: `ComponentCompilerMeta`, `CompilerCtx`, `BuildCtx`

packages/core/src/components/utils/modal/loading.ts
```diff
@@ -28,42 +32,56 @@ export type ModalLoadingOptions = {
...
-        modal.closeModal(null);
-        setTimeout(
-          async () => await getCoreDelegate().removeView(modal),
-          Animation.mediumTime
-        );
-      }, timeout);
-    },
-  };
+    return {
+      update: (text: string) => (loading.innerHTML = text),
```
dmytro-halimov: shouldn't we use `.innerText` here as well?

dmytro-halimov (changes requested):

packages/core/src/components/utils/dependency-function.ts
danielleroux: <strike>Yes that on purpose
``` 
  for (const dependency of dependencies) {
    await dependency.define();
  }

  await Promise.all(
    dependencies.map((dependency) => registry.whenDefined(dependency.tag))
  );
```
Line 51
</strike> 

Let me think about it you might be right

packages/core/src/components/utils/modal/loading.ts
danielleroux: Didn't challenged the existing implement here. Any thoughts?

packages/core/src/components/utils/dependency-function.ts
danielleroux: I had a wrong assumption that I need to have enumerable: true, but your right we can have it consistant not enumerable like the others.

packages/core/src/components/utils/modal/loading.ts
dmytro-halimov: innerText is just a safer option (excludes possible XSS injections), and since it's already used in this component, thought that maybe it makes sense to be consistent with it's usage.

dmytro-halimov (dismissed):

packages/core/src/components/utils/modal/loading.ts
danielleroux: Lets fix this
```
      finish: (text?: string, timeout: number = 250) => {
        if (text !== undefined) {
          loading.innerText = text; // <=== finish already using this 
        } else {
          timeout = 0;
```

danielleroux: Fixed thanks!

dmytro-halimov (approved):

PR #2544, opened 2026-05-12
fix(core/breadcrumb): Fix focus outline issues #IX-4265

dmytro-halimov:
JIRA ref: IX-4265

1. Double focus outline – navigate to the breadcrumb item, that has collapsed dropdown, click Enter to open it – focus stays on both the breadcrumb and first nested item
2. Focus line for first breadcrumb item is visually clipped (not only due to preview window in example)

1. After opening the dropdown, focus now is placed on the first dropdown item following standard accessibility guidelines.
2. Focus outline is now correctly visible at host edges, overflow: hidden was removed from the host

## 🏁 Checklist


packages/core/src/components/dropdown-button/dropdown-button.scss
```diff
@@ -134,6 +134,10 @@
   outline-offset: var(--theme-btn--focus--outline-offset);
 }
 
+:host([aria-expanded='true']:focus-visible) {
+  outline: none;
+}
+
```
danielleroux, line 140: I would prefer to combine it with the selector above, and do something like `:not(..selector...):focus-visible`

danielleroux (changes requested):

packages/core/src/components/dropdown-button/dropdown-button.scss
dmytro-halimov: I've tried this approach, but it has a side-effect: when the host is focused and expanded, the browser default focus outline becomes visible. Keeping the explicit `outline: none;` is necessary to suppress that default browser outline

danielleroux (approved):

PR #2545, opened 2026-05-12
fix(select): correct listbox screen reader selection and option role IX-4263

alexkaduk:

- **`ix-select`** exposes list semantics to assistive technologies via a **focus proxy** (`role="listbox"` / `role="option"`) and **`aria-activedescendant`** on the combobox input. **`FocusVisibleMixin`** was writing **`aria-selected="true"`** on **`ix-select-item`** whenever the row had keyboard focus visibility, and the proxy copied that attribute. Screen readers therefore announced **every focused option as “selected”**, even when it was not part of **`value`** (Firefox / Safari). 

- **`FocusVisibleMixin`** respects **`disableAriaSelectHandling`**: when **`true`**, the host **`aria-selected`** attribute is **removed** instead of mirroring **`ixFocusVisible`**. **`ix-select-item`** sets **`disableAriaSelectHandling={true}`** on its host so roving focus no longer implies selection.
- **`ix-select`** proxy options set **`aria-selected`** from **real state** (**`item.selected`** on **`ix-select-item`**, **`checked`** on the editable add row). **`aria-checked`** is no longer set on proxy options. Proxy **`aria-label`** falls back to **`value`** when **`label`** is missing (after author **`aria-label`** / **`label`**).
- **`ix-dropdown-item`** adds **`itemRole`** (default **`menuitem`**); **`ix-select-item`** passes **`itemRole="option"`** so the row surface matches listbox usage. Stencil output regenerates **Angular / React / Vue** bindings and **`components.d.ts`**.

## 🏁 Checklist

- [ ] 🗺️ Internationalization (i18n) - no hard coded strings — *N/A*
- [ ] 📕 Add or update a Storybook story — *N/A unless you extend select a11y stories*
- [ ] 📄 Documentation was reviewed/updated [siemens/ix-docs](https://github.com/siemens/ix-docs) — *API text follows updated JSDoc on `itemRole` / mixin after usual docs sync if required*
- [ ] 🧪 Unit tests were added/updated and pass (`pnpm test`) — *+2 CT cases in `select.ct.ts`; run `pnpm test` or at least `packages/core` spec + CT*
- [ ] 📸 Visual regression tests were added/updated and pass ([Guide](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md#visual-regression-testing)) — *N/A (no intentional pixel change)*

## 👨‍💻 Help & support

- Re-verify with **VoiceOver** (Chrome / Safari) and keyboard: open **`ix-select`**, arrow through options with **no** or **partial** selection; confirm labels, position, and **“selected”** only on values in **`value`**.
- **`itemRole`** is additive with default **`menuitem`**; existing **`ix-dropdown-item`** usages keep prior behavior unless they set **`item-role`**.

### DIFF
--- a/packages/core/src/components/dropdown-item/dropdown-item.tsx
+++ b/packages/core/src/components/dropdown-item/dropdown-item.tsx
@@ -32,6 +32,8 @@ import {
 } from '../utils/internal/mixins/id.mixin';
 import { FocusVisibleMixin } from '../utils/internal/mixins/focus-visible.mixin';
 
+export type IxDropdownItemRole = 'menuitem' | 'option';
+
 @Component({
   tag: 'ix-dropdown-item',
   styleUrl: 'dropdown-item.scss',
@@ -84,10 +86,12 @@ export class DropdownItem
   @Prop({ reflect: true }) checked = false;
 
   /**
-   * Role of the host surface (`menuitem` or `option`).
-   * Use `option` when the item represents a listbox option (e.g. inside select).
+   * Role of the host surface.
+   * Use `option` when the item represents a listbox option (e.g. inside select); use `menuitem` in menus.
+   *
+   * @since 5.0.0
    */
-  @Prop() itemRole: string = 'menuitem';
+  @Prop() itemRole: IxDropdownItemRole = 'menuitem';
 
   /** @internal */
   @Prop() isSubMenu = false;
--- a/packages/core/src/components/dropdown-item/dropdown-item.tsx
+++ b/packages/core/src/components/dropdown-item/dropdown-item.tsx
@@ -31,8 +31,7 @@ import {
   ComponentIdMixinContract,
 } from '../utils/internal/mixins/id.mixin';
 import { FocusVisibleMixin } from '../utils/internal/mixins/focus-visible.mixin';
-
-export type IxDropdownItemRole = 'menuitem' | 'option';
+import type { IxDropdownItemRole } from './dropdown-item.types';
 
 @Component({
   tag: 'ix-dropdown-item',
--- a/packages/core/src/components/dropdown-item/dropdown-item.tsx
+++ b/packages/core/src/components/dropdown-item/dropdown-item.tsx
@@ -83,6 +83,12 @@ export class DropdownItem
    */
   @Prop({ reflect: true }) checked = false;
 
+  /**
+   * Role of the host surface (`menuitem` or `option`).
+   * Use `option` when the item represents a listbox option (e.g. inside select).
+   */
+  @Prop() itemRole: string = 'menuitem';
+
   /** @internal */
   @Prop() isSubMenu = false;
 
@@ -130,7 +136,7 @@ export class DropdownItem
     return (
       <Host
         id={id}
-        role={'menuitem'}
+        role={this.itemRole}
         aria-disabled={a11yBoolean(this.disabled)}
         aria-label={this.hostElement.ariaLabel ?? this.ariaLabelButton}
         class={{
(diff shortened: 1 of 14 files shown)


packages/core/src/components/dropdown-item/dropdown-item.tsx
```diff
@@ -32,6 +32,8 @@ import {
 } from '../utils/internal/mixins/id.mixin';
 import { FocusVisibleMixin } from '../utils/internal/mixins/focus-visible.mixin';
 
+export type IxDropdownItemRole = 'menuitem' | 'option';
```
danielleroux: Do not expose the type from the component file. Use a `.types.ts` -file and export it via the public-api.ts

packages/core/src/components/select/select.tsx
```diff
@@ -208,6 +208,13 @@ export class Select
    */
   @Prop({ attribute: 'i18n-all-selected' }) i18nAllSelected = 'All';
 
+  /**
+   * Prefix for the accessible name of the close control on a selected chip in multiple mode.
+   * The chip label or value is appended (e.g. "Remove Item 1").
+   */
+  @Prop({ attribute: 'i18n-remove-selected-item' }) i18nRemoveSelectedItem =
```
danielleroux, line 217: `since`annotation is missing

packages/core/src/components/select/select.tsx
```diff
@@ -828,10 +844,17 @@ export class Select
...
+          item.label ||
+          (item.tagName === 'IX-SELECT-ITEM'
+            ? (item as HTMLIxSelectItemElement).value
+            : '') ||
+          '';
+        const selected =
+          item.tagName === 'IX-SELECT-ITEM'
+            ? (item as HTMLIxSelectItemElement).selected
+            : (item as HTMLIxDropdownItemElement).checked;
+        proxyElement.ariaSelected = a11yBoolean(!!selected);
```
danielleroux, line 859: Maybe not the best to put this all inline? Maybe make the code more readable and put it into a variable and assign it then

packages/core/src/components/utils/internal/mixins/focus-visible.mixin.ts
```diff
@@ -28,7 +28,11 @@ export const FocusVisibleMixin = <B extends MixedInCtor<StencilLifecycle>>(
     @Prop({ reflect: true }) ixFocusVisible = false;
 
     /**
+     * When `true`, do not map keyboard focus visibility to `aria-selected` on the host.
+     * Use when selection state must not mirror roving focus (e.g. `ix-select-item`).
+     *
      * @internal
+     * @since 5.0.0
```
danielleroux: Internal api does not need since annotation

danielleroux (changes requested):

packages/core/src/components/dropdown-item/dropdown-item.tsx
alexkaduk: moved. thanks

packages/core/src/components/select/select.tsx
alexkaduk: added. thanks

packages/core/src/components/utils/internal/mixins/focus-visible.mixin.ts
alexkaduk: updated. thanks

packages/core/src/components/select/select.tsx
alexkaduk: refactored. thanks

danielleroux (approved):

PR #2573, opened 2026-05-28
feat(core/menu): Implement w3c menu pattern for ix-menu

lzeiml:
IX-4304

Currently ix-menu does not have proper keyboard-navigation according to w3c menubar pattern: https://www.w3.org/WAI/ARIA/apg/patterns/menubar/examples/menubar-navigation/

Implements the w3c keyboard-navigation and roles for better screenreader support.

## 🏁 Checklist

## Summary by CodeRabbit

* **New Features**
  * Added accessibility i18n props for `ix-menu` and enhanced menubar support with roving `tabIndex`, improved Arrow/Home/End navigation, and utility controls grouping.
  * Introduced public `setTabIndex()` APIs for menu items/categories and improved category labeling support; added typings for `ix-playground`.
* **Bug Fixes / Accessibility**
  * Refined dropdown open/close and focus restore behavior, updated ARIA roles/attributes (e.g., tooltip, menu item state), and improved keyboard activation handling.
  * Updated expand/collapse icons to use `aria-expanded`.
* **Tests**
  * Updated Playwright and visual e2e tests to match updated `menuitem` roles and interactions.
* **Chores**
  * Refreshed public typings (including Angular proxy inputs) and added an accessibility patch changeset.


lzeiml:
~~remaining sonar issues are not relevant:~~
~~- menubar does not have to be focusable directly, at least according to W3C~~
~~- group role is needed to satisfy axe a11y check~~
~~- tabIndex on div is needed to implement desired scrolling behaviour when a lot of items are present~~

packages/core/src/components/menu-item/menu-item.tsx
```diff
@@ -177,15 +215,50 @@ export class MenuItem implements IxMenuItemBase {
...
+      return this.inheritAriaAttributes['aria-label'];
+    }
+
+    const hasDistinctTooltip =
+      this.tooltipText &&
+      this.tooltipText !== this.label &&
+      this.tooltipText !== this.hostElement.textContent;
+
+    if (hasDistinctTooltip) {
+      return `${this.label ?? this.menuCategoryLabel} ${this.tooltipText}`;
```
dmytro-halimov: it looks like slotted menu items with a custom tooltip can end up with something like "undefined custom tooltip text".

for example, in case like this one:
`<ix-menu-item tooltip-text="my tooltip">Foo bar</ix-menu-item>`

`hasDistinctTooltip` correctly compares against `this.hostElement.textContent`, but the returned label only falls back to `this.label ?? this.menuCategoryLabel`.
Could we include `this.hostElement.textContent` in that fallback?

packages/core/src/components/menu/menu.tsx
```diff
@@ -678,93 +818,107 @@ export class Menu {
...
+                    'tabs--shadow-top': true,
+                    'tabs--shadow--show': this.itemsScrollShadowTop,
+                  }}
+                ></div>
+                <div
+                  class={{
+                    tabs: true,
+                    'show-scrollbar': this.expand,
+                  }}
+                  role="group"
```
dmytro-halimov: should we consider adding an aria-label for parent context?
Without aria-label this div with role="group" is announced just as an unnamed group, without giving the context in which menu you are currently in

packages/core/src/components/menu/menu.tsx
```diff
@@ -617,6 +623,140 @@ export class Menu {
...
+    allMenuItems.forEach((item) => {
+      item.removeAttribute('aria-posinset');
+      item.removeAttribute('aria-setsize');
+    });
+
+    const total = items.length;
+
+    items.forEach((item, index) => {
+      item.setAttribute('aria-posinset', String(index + 1));
+      item.setAttribute('aria-setsize', String(total));
```
dmytro-halimov, line 738: I've noticed that screen readers don't announce amount of items in menu (like 2nd of 4). Probably the issue is thatfor categories, item is the `ix-menu-categor`y host, which has no role. The role="menuitem" is on the inner `.category-parent` button.

I think we need to delegate it to the inner item like `setTabIndex` already does in menu-category component, instead of `setAttribute` on the host.

dmytro-halimov (changes requested):

packages/core/src/components/menu/menu.tsx
lzeiml: turns out those roles were not needed after all so i removed them

packages/core/src/components/menu/menu.tsx
lzeiml: i applied the missing attributes for menu-item inside menu-category, should work now

packages/core/src/components/menu-item/menu-item.tsx
lzeiml: good catch, added hostElement.textContent as additional fallback and empty string as last fallback to avoid undefined

lzeiml:
regarding open sonar issues:
aria-description does in fact work, and it is not flagged by storybook.
`<dialog>` instead of `<section>` breaks the overlay

.changeset/fair-falcons-search.md
```diff
@@ -0,0 +1,5 @@
+---
+'@siemens/ix': patch
```
danielleroux: Introduce new properties is no "patch" is actual a "minor" change

testing/visual-testing/tests/menu/link/index.html
```diff
@@ -18,22 +18,12 @@
...
-          </a>
-          <a href="#link3">
-            <ix-menu-item active>Link 3</ix-menu-item>
-          </a>
-          <a href="#link4">
-            <ix-menu-item>Link 4</ix-menu-item>
-          </a>
-          <a href="#link5">
-            <ix-menu-item>Link 5</ix-menu-item>
-          </a>
```
danielleroux, line 36: Why do you modify the test here? Wrap the menu in a[href] is valid scenario to use.

danielleroux (changes requested):

testing/visual-testing/tests/menu/link/index.html
lzeiml: I changed this because i thought this was a deprecated leftover, my bad. it is restored now

packages/core/src/components/menu/menu.tsx
```diff
@@ -566,9 +605,39 @@ export class Menu {
...
+  private focusMenuNavigationContainer() {
+    const shadowRoot = this.hostElement.shadowRoot;
+
+    if (!shadowRoot) {
+      return;
+    }
+
+    const menuNavigation = shadowRoot.querySelector(
+      '.menu-navigation'
+    ) as HTMLDivElement | null;
```
nuke-ellington: We already have a getter for this (`this.menuNaviationContainer`)

packages/core/src/components/menu-category/test/menu-category.ct.ts
```diff
@@ -390,3 +391,120 @@ test('should adjust height when items are added dynamically', async ({
...
+    const anchor1 = page.locator('#cat-anchor1');
+    const anchor2 = page.locator('#cat-anchor2');
+
+    await expect(anchor1).toHaveAttribute('tabindex', '-1');
+    await expect(anchor2).toHaveAttribute('tabindex', '-1');
+  }
+);
+
+regressionTest(
+  'show category if anchor-wrapped items are focused',
```
nuke-ellington: Test name is confusing

nuke-ellington (changes requested):

nuke-ellington (approved):

PR #2574, opened 2026-05-31
feat(popover): implement popover component IX-3942

alexkaduk:
- IX has no dedicated **popover** component for anchored floating panels.

- Adds **`ix-popover`** (`@since 5.1.0`) as a floating panel anchored to a trigger via Floating UI.
- Adds sub-components: **`ix-popover-header`**, **`ix-popover-content`**, **`ix-popover-image`**, **`ix-popover-footer`** (slots, paddingless content, vertical footer alignment, optional header icon/close).
- **Triggers:** `trigger` (selector or element ref), `triggerMode` `click` | `hover`, programmatic **`showPopover()`** / **`hidePopover()`**, and reflected **`show`** with cancelable **`showChange`**.
- **Interaction:** optional `closeOnClickOutside`, Escape, header close (`closeClick`), nesting via `ix-assign-sub-popover` and global **`popoverController`** (aligned with dropdown patterns).
- **Accessibility:** `<dialog popover="manual">`, trigger `aria-expanded` / `aria-controls` / `aria-haspopup="dialog"`, focus trap when focusable content exists, `aria-modal` for interactive panels, host `aria-label` / custom `role` forwarding.
- **Refactor:** extracts **`NestedOverlayStack`** under `components/utils/nested-overlay/` and wires **dropdown** + **popover** controllers to it (DRY overlay hierarchy / outside-click dismiss).

## 🏁 Checklist

## Summary by CodeRabbit

* **New Features**
  * Introduced ix-popover with header/content/image/footer slots, click/hover triggers, placement options, optional spike, nesting, and programmatic open/close.
* **Accessibility / UX**
  * Improved focus management and keyboard handling for overlays and nested popovers.
* **Documentation / Examples**
  * Added Angular/React/Vue preview examples and Storybook demos.
* **Tests**
  * Added extensive unit, integration, Playwright, visual and accessibility tests covering interactions, nesting, focus, and ARIA behavior.


packages/core/scss/theme/core/_components.scss
danielleroux, line 14: Just a hint that you read this here. Because adaption of the theme requires a theme generator update (not a public repo)

packages/core/src/components/popover-header/popover-header.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * Icon color
+   */
+  @Prop() iconColor?: string;
+
+  /**
+   * ARIA label for the icon
+   */
+  @Prop() ariaLabelIcon?: string;
```
danielleroux: Do we need in this case a label for the icon? There is the h5 which should give enough context

packages/core/src/components/popover/popover-controller.ts
```diff
@@ -0,0 +1,203 @@
...
+
+/**
+ * Focus management strategy when an interactive popover closes.
+ *
+ * - `restore-trigger`: Move focus back to the trigger element (default, WCAG-compliant).
+ * - `release`: Blur the active element without restoring focus (hover mode pointer dismiss).
+ */
+export type PopoverCloseFocus = 'restore-trigger' | 'release';
+
+export interface PopoverInterface {
```
danielleroux, line 23: I would prefer to refactor then the dropdown-controller interface also, this this is typical the same implementation with some slide differences, also the implementation is similar to the dropdown-controller implementation.

e.g

PopoverInterface extends DropdownInterface or vice visa

packages/core/src/components/popover/popover.scss
danielleroux, line 1: Use the `:host` selector

packages/core/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,759 @@
...
+
+    if (newValue) {
+      void this.openPopover({ fromShowWatch: true });
+    } else {
+      void this.closePopover({ fromShowWatch: true });
+    }
+  }
+
+  componentWillLoad() {
+    this.ariaAttributes = a11yHostAttributes(this.hostElement);
```
danielleroux: Use the mixin for this case

packages/core/src/components/popover/test/popover.constants.ts
```diff
@@ -0,0 +1,21 @@
...
+ * SPDX-License-Identifier: MIT
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+/**
+ * Hover hide delay in milliseconds (matches implementation)
+ */
+export const HOVER_HIDE_MS = 200;
```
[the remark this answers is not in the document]
alexkaduk: adjusted

packages/core/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,823 @@
...
+    return (
+      <Host class={{ visible: this.show }} data-ix-popover={this.uid}>
+        <dialog
+          ref={this.dialogRef}
+          id={this.uid}
+          class="dialog"
+          popover="manual"
+          inert={!this.show}
+          role={role}
+          aria-modal={a11yBoolean(this.hasFocusableContent)}
```
danielleroux: I think aria-modal="true" if there is an focusable element is inside is not 100% correct because lets check following page: https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/examples/dialog/#rps_label

Image


> Tells assistive technologies that the windows underneath the current dialog are not available for interaction (inert).

But they are intractable

packages/core/src/components/popover/popover.tsx
danielleroux, line 1: Accessing nested popover via keyboard is possible but once reached the nested one its not possible to focus the other popover again without closing all popover (ESC)

```
        <button ref={this.ref}>Trigger</button>
        <ix-popover trigger={this.ref.waitForCurrent()}>
          <button>1</button>
          <button>2</button>
          <button>3</button>
          <button>4</button>
          Parent: <button ref={this.ref2}>Trigger2</button>
        </ix-popover>

        <ix-popover trigger={this.ref2.waitForCurrent()}>
          <button>3</button>
          <button>4</button>Inner
        </ix-popover>
```

Im not sure if this is a real accessibility issue or just a QOL feature to come back to the underlaying popover

danielleroux, line 1: Spike overlays content:


Image


Code:

```
        <button ref={this.ref}>Trigger</button>
        <ix-popover trigger={this.ref.waitForCurrent()}>
          <button>1</button>
          <button>2</button>
          <button>3</button>
          <button>4</button>
          Parent: <button ref={this.ref2}>Trigger2</button>
        </ix-popover>

        <ix-popover trigger={this.ref2.waitForCurrent()} hasSpike>
          <button>3</button>
          <button>4</button>Inner
        </ix-popover>
```

danielleroux (changes requested):

packages/core/src/components/popover/popover.tsx
danielleroux, line 1:
```
@Component({
  tag: 'ix-comp-a',
  shadow: true,
})
export class CompA {
  private ref = makeRef<HTMLButtonElement>();
  private ref2 = makeRef<HTMLButtonElement>();
  render() {
    return (
      <Host>
        <button ref={this.ref}>Trigger</button>
        <ix-popover trigger={this.ref.waitForCurrent()}>
          <button>1</button>
          <button>2</button>
          <button>3</button>
          <button>4</button>
          Parent: <button ref={this.ref2}>Trigger2</button>
        </ix-popover>
        <ix-popover trigger={this.ref2.waitForCurrent()} hasSpike>
          <button>3</button>
          <button>4</button>Inner
        </ix-popover>
      </Host>
    );
  }
}
```

Popover inside shadowDOM focus trap not working like outside of the shadowDOM.
Snippet above does not trap any focus at all.

danielleroux (changes requested):

packages/core/scss/theme/core/_components.scss
alexkaduk: Will remove/update it when theme is updated/generated.

packages/core/src/components/popover-header/popover-header.tsx
alexkaduk: ddrop ariaLabelIcon

header icon is decorative whenever a title is present

packages/core/src/components/popover/popover-controller.ts
alexkaduk: will be handled later with separate PR

packages/core/src/components/popover/popover.scss
alexkaduk: nest styles under :host and add :host(.visible) to match the existing host class. no behavior change intended, just shadow-DOM structure

packages/core/src/components/popover/popover.tsx
alexkaduk: switched to InheritAriaAttributesMixin + DefaultMixins. 
host ARIA is forwarded to the dialog via inheritAriaAttributes instead of a manual componentWillLoad call

packages/core/src/components/popover/popover.tsx
alexkaduk: agreed, aria-modal="true" only applies when the page outside is inert and obscured. 

omit aria-modal on the dialog

packages/core/src/components/popover/popover.tsx
alexkaduk: the topmost layer should trap focus. it was adjusted. now Escape dismisses only one level so users can return to the parent without closing the whole stack

implemented LIFO Escape and layered focus traps (parent defers when focus is in the child). Tab back into the parent while the child is still open is intentionally not supported

packages/core/src/components/time-picker/time-picker.tsx
nuke-ellington, line 1: Why not use the constants for the number properties anymore?

packages/core/src/components/popover/popover.tsx
alexkaduk: fixed. thanks!

alexkaduk: fixed. thanks
shadow DOM broke the trap because I compared against document.activeElement (shadow host) instead of the deep active element inside the shadow tree. nested inner popover also needed the parent trap to defer when it isn’t topmost

packages/core/src/components/popover/popover-controller.ts
```diff
@@ -0,0 +1,216 @@
...
+  async presentAndWait(popover: PopoverInterface): Promise<void> {
+    if (!popover.isPresent() && popover.willPresent?.()) {
+      this.stack.forEach((openPopover) => {
+        if (openPopover.isPresent()) {
+          this.stack.setChildIds(
+            openPopover.getId(),
+            openPopover.getNestedPopoverIds()
+          );
+        }
+      });
```
nuke-ellington: Is this block really necessary? Should the stack not already reflect the current popover hierarchy?

packages/core/src/components/popover-content/popover-content.tsx
```diff
@@ -0,0 +1,46 @@
...
+})
+export class PopoverContent {
+  @Element() hostElement!: HTMLIxPopoverContentElement;
+
+  /**
+   * Remove default inner padding
+   *
+   * @since 5.1.0
+   */
+  @Prop() paddingless = false;
```
nuke-ellington: Other components use `noPadding` as property name. We should keep this consistent.

packages/core/src/components/popover-footer/popover-footer.scss
```diff
@@ -0,0 +1,44 @@
...
+  .footer-start {
+    display: flex;
+    align-items: center;
+    gap: 0.5rem;
+  }
+
+  .footer-end {
+    display: flex;
+    align-items: center;
+    gap: 0.5rem;
```
nuke-ellington: Redundant (also with host)

packages/core/src/components/popover-image/popover-image.tsx
```diff
@@ -0,0 +1,49 @@
...
+   */
+  @Prop() src?: string;
+
+  /**
+   * Alt text for the image.
+   * Use an empty string for decorative images; provide descriptive text for content images.
+   *
+   * @since 5.1.0
+   */
+  @Prop() alt = '';
```
nuke-ellington: Both `src` and `alt` are HTML attributes. We try to avoid these as prop names.

packages/core/src/components/utils/focus/focus-trap.ts
nuke-ellington, line 1: There is a lot of DOM traversal happening how, which might affect performance negatively.

nuke-ellington (dismissed):

packages/core/src/components/time-picker/time-picker.tsx
alexkaduk: to avoid such error on docu https://ix.siemens.io/docs/components/time-picker/code#property-hourInterval
image
check `Default` row

packages/core/src/components/popover/popover-controller.ts
alexkaduk: agree — the stack should reflect hierarchy without re-syncing every open popover on each present. the gap was that childIdsByParent only updated on setChildIds, while hosts learn nested IDs via ix-assign-sub-popover (which can happen after the parent is already open)

moved that sync into the ix-assign-sub-popover path (syncNestedPopoverIds) so parent→child links stay current when nested popovers register, and removed the blanket forEach in presentAndWait. dismissOthers() now relies on that incremental sync plus a single sync for the popover being opened

added a CT that injects the nested popover after the parent opens to guard timing

packages/core/src/components/popover-content/popover-content.tsx
alexkaduk: agreed — renamed `paddingless` → `noPadding` to match ix-pane / ix-date-time-card

HTML attribute is no-padding; host class updated accordingly. tests and storybook updated

thanks!

packages/core/src/components/popover-footer/popover-footer.scss
alexkaduk: removed `margin-inline-start: auto` on .footer-end since. 
kept `display: flex + gap` on the inner wrappers so multiple slotted buttons / start items still group correctly; `:host` gap only spaces the two regions

packages/core/src/components/popover-image/popover-image.tsx
alexkaduk: agreed. thanks!
renamed `src/alt` to `image/imageAlt` to avoid HTML attribute prop names (same pattern as ix-avatar). 
maps to `image / image-alt` in markup; still renders native `src/alt` on the inner

packages/core/src/components/utils/focus/focus-trap.ts
alexkaduk: agree. at the same time the traversal runs only on Tab, not continuously. popover needs it for top-layer + ancestor shadow DOM and nested popovers — native focus trap doesn’t handle that. scope is one keydown handler and a small focusable list per open popover

packages/core/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,826 @@
...
+    if (!this.hasFocusableContent || !this.triggerElement) {
+      return;
+    }
+
+    if (this.triggerMode === 'hover' && this.closeFocus === 'release') {
+      this.releasePopoverFocus();
+      return;
+    }
+
+    this.triggerElement.focus();
```
danielleroux: This should only be executed if the dropdown is closed via keyboard (`hasKeyboardMode()`) which should give context about if keyboard is used otherwise if you click outsite it steels your focus and show the outline effect

packages/core/src/components/popover/popover.tsx
danielleroux, line 1: Differences regarding focusing inside the dropdown

### Keyboard
- Navigate to trigger element, and open popover via keyboard
- Press Tab first element is focused

### Mouse
- Click on trigger element, to open popover
- Press Tab second element is focused

I would expect that Tab after Mouse interaction also focus first focusable element

packages/core/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,826 @@
...
+
+  private clearHideTimeout() {
+    if (this.hideTimeout !== undefined) {
+      clearTimeout(this.hideTimeout);
+      this.hideTimeout = undefined;
+    }
+  }
+
+  private getTriggerAriaTarget(): HTMLElement {
+    const el = this.triggerElement!;
```
danielleroux:
```suggestion
    const el = this.triggerElement;
```

Because trigger element could go to null/undefined

```
  private triggerElement?: HTMLElement;
```

danielleroux (changes requested):

packages/core/src/components/popover/popover.tsx
alexkaduk: removed the non-null assertion; getTriggerAriaTarget() now handles optional triggerElement explicitly

packages/core/src/components/popover/popover.tsx
alexkaduk: fixed. thanks

packages/core/src/components/popover/popover.tsx
```diff
@@ -390,9 +398,18 @@ export class Popover
...
+      focusOptions
+    );
     requestAnimationFrame(() => {
-      focusFirstFocusTrapElement(this.hostElement, focusTrapOptions);
+      focusFirstFocusTrapElement(
+        this.hostElement,
+        focusTrapOptions,
+        focusOptions
+      );
     });
```
danielleroux: Is the double call here really necessary? Other point avoid the requestAnimationFrame use the NoZone variant to avoid ngZone trigger

packages/core/src/components/utils/focus/focus-trap.ts
nuke-ellington: @danielleroux thoughts?

packages/core/src/components/utils/focus/focus-trap.ts
```diff
@@ -17,125 +23,516 @@ export interface FocusTrapResult {
...
+      handled = true;
+    }
 
+    if (handled) {
       event.stopImmediatePropagation();
     }
   };
 
-  ref.addEventListener('keydown', onKeydown);
+  listenerTarget.addEventListener('keydown', onKeydown, true);
```
danielleroux: Using "capture" = `true` will prevent the underlaying elements in the tree to listen for keyboard events = "tabs".

```
<ix-popover>
    <ix-button onKeyDown={() => console.log('show every keydown but not Tab key')}
    <button>test</button>
```

e.g if using ix-time-picker inside popover will prevent this case to work: https://github.com/siemens/ix/blob/fa6a6aae17c53f7718d84098454a3a4ad7b94dad/packages/core/src/components/time-picker/time-picker.tsx#L482-L486

packages/core/src/components/utils/focus/focus-trap.ts
danielleroux, line 1: Didnt checked the code before the refactoring regarding this bug. If one element in the list is not focusable e.g if it `display: 'none'` or have `inert` attribute the focus trap is stuck there.

With the following snippet you can never "jump" over button 2
```
          <ix-popover>
            <button>1</button>
            <button style={{ display: 'none' }}>2</button>
            <button>3</button>
            <button>4</button>
            <button data-popover-close>5</button>
          </ix-popover>
```

danielleroux: Maybe some review regarding angular and ngZone is required which can cause "change detections" (@alexkaduk maybe just a short check is enough). In case that this is only happing with Tab key I assume this is also not a bug deal, but lets spend some manual effort and stress test this

packages/core/src/components/utils/focus/focus-trap.ts
```diff
@@ -17,125 +23,516 @@ export interface FocusTrapResult {
 export const TRAP_FOCUS_EXCLUDE_ATTRIBUTE = 'data-ix-focus-trap-exclude';
 export const TRAP_FOCUS_INCLUDE_ATTRIBUTE = 'data-ix-focus-trap-include';
 
+const POPOVER_TAG = 'IX-POPOVER';
+const POPOVER_HEADER_TAG = 'IX-POPOVER-HEADER';
```
danielleroux, line 27: focus-trap has now some kind of dependency to popover, is this really necessary? Lets discuss this first no direct code adaption needed

packages/core/src/components/utils/focus/focus-trap.ts
```diff
@@ -17,125 +23,516 @@ export interface FocusTrapResult {
...
+    const activeIndex = findActiveFocusableIndex(
+      activeElement,
+      focusableElements
+    );
 
-      if (focusableElements.length === 0) {
+    if (options?.listenOnDocument) {
+      if (options.shouldDeferTabTrap?.(trapHost)) {
         return;
       }
```
danielleroux: Can we not just move the `if (options.shouldDeferTabTrap?.(trapHost))` before the "expensive" call of `getFocusTrapFocusables:443`. Because if the Tab is not event interesting for the host then we can skip it. (not tested locally)

packages/core/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,890 @@
...
+          trap.destroy();
+        } else {
+          this.focusTrap = trap;
+          if (this.triggerMode !== 'hover') {
+            this.focusFirstElement();
+          }
+        }
+      }
+
+      requestAnimationFrame(() => {
```
danielleroux: Lets use the `requestAnimationFrameNoZone` to remove the ngZone detect cycle

danielleroux (changes requested):

packages/core/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,890 @@
...
+      <Host class={{ visible: this.show }} data-ix-popover={this.uid}>
+        <dialog
+          ref={this.dialogRef}
+          id={this.uid}
+          class="dialog"
+          popover="manual"
+          inert={!this.show}
+          role={role}
+          aria-label={this.inheritAriaAttributes['aria-label']}
+          aria-describedby={this.inheritAriaAttributes['aria-describedby']}
```
danielleroux: Why do you apply only the aria-label and aria-describedby, I would suggest to pass through all aria labels.

```
<dialog
    // other attributes
    {...this.inheritAriaAttributes}
```

packages/core/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,890 @@
...
+})
+export class Popover
+  extends Mixin(...DefaultMixins, InheritAriaAttributesMixin)
+  implements PopoverInterface, InheritAriaAttributesMixinContract
+{
+  @Element() override hostElement!: HTMLIxPopoverElement;
+
+  /**
+   * Element that toggles the popover.
+   * CSS selector string or DOM element reference.
```
danielleroux: `findElement` just check for IDs not css selectors like #trigger this .trigger won’t work

danielleroux (changes requested):

packages/core/src/components/popover/popover.tsx
alexkaduk: adjusted. thanks

packages/core/src/components/utils/focus/focus-trap.ts
alexkaduk: there was issue for Tab moves hour → minute → second inside the popover. fixed
the trap only wraps at listed controls (Apply/header Close), not on every Tab from shadow-DOM picker cells

packages/core/src/components/utils/focus/focus-trap.ts
alexkaduk: filter CSS-hidden elements out when building the trap list (isHiddenFromTabOrder on computed display / visibility). Inert / nested-popover exclusion stays in shouldExcludeFromFocusTrap as before.

added tests; verified manually on the React preview (hidden + visibility: hidden cases).

packages/core/src/components/utils/focus/focus-trap.ts
alexkaduk: the util currently knows ix-popover / ix-popover-header for nested traps, top-layer deferral, and “focus content not header”. that’s popover-only today

dropdown/pane use the generic path. they don’t pass listenOnDocument and don’t rely on the tag checks unless nested popovers appear in content

Possible decoupling:

1) move popover-specific bits to e.g. popover/popover-focus-trap.ts (wrapper or options factory), keep focus-trap.ts generic.
2) or extend FocusTrapOptions with hooks: getOverlayHost, shouldExcludeFocusable, preferFocusTarget — popover supplies implementations, util stays tag-agnostic

3) or a neutral marker (e.g. data-ix-overlay-host) instead of IX-POPOVER / IX-POPOVER-HEADER tag names

packages/core/src/components/utils/focus/focus-trap.ts
alexkaduk: applied

shouldDeferTabTrap and nested-layer check now run before getFocusTrapFocusables. 

added a unit test for nested-layer deferral

packages/core/src/components/popover/popover.tsx
alexkaduk: adjusted

packages/core/src/components/popover/popover.tsx
alexkaduk: adjusted

packages/core/src/components/popover/popover.tsx
alexkaduk: clarified trigger prop accepts id, not CSS selector

packages/angular-standalone-test-app/src/preview-examples/popover.ts
```diff
@@ -0,0 +1,77 @@
...
+        Check out the new dashboard and improved performance metrics.
+      </ix-popover-content>
+      <ix-popover-footer alignment="vertical">
+        <span slot="start">v4.0.0</span>
+        <ix-button variant="secondary" (click)="popover.hidePopover()"
+          >Cancel</ix-button
+        >
+        <ix-button>Save</ix-button>
+      </ix-popover-footer>
+    </ix-popover>
```
danielleroux: For angular we try to make standalone html templates. Its easier to read inside the playground in ix.siemens.io.

packages/angular-test-app/src/preview-examples/popover.ts
```diff
@@ -0,0 +1,55 @@
...
+      </ix-popover-content>
+      <ix-popover-footer alignment="vertical">
+        <span slot="start">v4.0.0</span>
+        <ix-button variant="secondary" (click)="popover.hidePopover()"
+          >Cancel</ix-button
+        >
+        <ix-button>Save</ix-button>
+      </ix-popover-footer>
+    </ix-popover>
+  `,
```
danielleroux: Same as in standalone please use a template file

packages/core/src/components/popover-footer/popover-footer.tsx
```diff
@@ -0,0 +1,52 @@
...
+
+  /**
+   * Button layout direction
+   *
+   * @since 5.1.0
+   */
+  @Prop() alignment: 'horizontal' | 'vertical' = 'horizontal';
+
+  componentDidLoad() {
+    this.hostElement.setAttribute(TRAP_FOCUS_INCLUDE_ATTRIBUTE, '');
```
danielleroux: You can also use this inside the render function, because from lifecycle perspective [componentDidLoad once](https://stenciljs.com/docs/component-lifecycle) which mean any manipulation of the data attributes can remove the tag. (same for other components which have that style)

```
{...{ [TRAP_FOCUS_INCLUDE_ATTRIBUTE]: true }}
```

packages/core/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,896 @@
...
+  showChange!: EventEmitter<boolean>;
+
+  /**
+   * Fires after visibility has changed
+   *
+   * @since 5.1.0
+   */
+  @Event() showChanged!: EventEmitter<boolean>;
+
+  private readonly uid = `popover-${popoverInstance++}`;
```
danielleroux: Just a idea there exist a `ComponentIdMixin`, maybe we can utilize this one here?

packages/core/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,896 @@
...
+
+  private readonly assignedNestedPopoverIds: string[] = [];
+  private hasFocusableContent = false;
+  private suppressShowWatch = false;
+  private isOpeningPopover = false;
+  private closeFocus: PopoverCloseFocus = 'restore-trigger';
+  private hasDisconnected = false;
+  private triggerRegistryId = 0;
+  /** Hover mode: block focus handler from re-opening after programmatic dismiss focus. */
+  private suppressFocusPresent = false;
```
danielleroux: Is this a leftover comment?

packages/core/src/components/popover/popover.tsx
```diff
@@ -0,0 +1,896 @@
...
+  @Element() override hostElement!: HTMLIxPopoverElement;
+
+  /**
+   * Element that toggles the popover.
+   * String values are resolved as the trigger element `id`, not as CSS selectors.
+   * Also accepts a DOM element reference.
+   *
+   * @since 5.1.0
+   */
+  @Prop() trigger?: ElementReference;
```
danielleroux, line 106: Found a small issue:

```
<button ref={ref => this.buttonRef = ref>Test</button>
<ix-popover trigger={this.buttonRef}>
</ix-popover>
```

Which will render the ix-popover first with trigger null/undefined (I know via api not allowed, but this will solve so many issues if the issue is solved). If then later the trigger can set to a valid element, the trigger is still not working

packages/core/src/components/popover/popover.tsx
danielleroux, line 1: Currently the popover the slotted content of the popover is a flex layout by default. Not a big deal in that way, but I would expect to have a neutral/standard "div"-like setting there

danielleroux (changes requested):

packages/angular-standalone-test-app/src/preview-examples/popover.ts
alexkaduk: refactored. thanks

packages/angular-test-app/src/preview-examples/popover.ts
alexkaduk: refactored. thanks

packages/core/src/components/popover-footer/popover-footer.tsx
alexkaduk: updated footer, content, and header. thanks

packages/core/src/components/popover/popover.tsx
alexkaduk: used ComponentIdMixin, thanks for highlighting

packages/core/src/components/popover/popover.tsx
alexkaduk: not leftover — still needed for hover + Escape. programmatic refocus on the trigger would hit the hover focus handler and reopen the popover. the flag suppresses present() for one frame. clarified the comment to point at schedulePostCloseFocus() and the focus handler

packages/core/src/components/popover/popover.tsx
alexkaduk: added componentDidRender to register when trigger is set but not yet wired.
and tightened @Watch('trigger') to re-run on real changes

packages/core/src/components/popover/popover.tsx
alexkaduk: switched the slot wrapper to `display: block` so content behaves like a standard container.
header/footer keep internal flex for their own layout
thanks!

danielleroux (approved):

packages/core/src/components/utils/focus/focus-trap.ts
alexkaduk: Manually stress-tested Tab in `angular-standalone-test-app/preview/popover` with temporary
ngDoCheck logging (27+ Tab presses with popover open). 

Every Tab reported `doCheckDelta: 0` — no extra change detection cycles on the preview component.

I think we're good for this PR. Manual check didn't show a CD issue.

danielleroux (approved):

Issue #2595, opened 2026-06-10
Ix modal can not be closed correctly


chencc1988:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

Sometimes, modal windows need to be closed using Redux events. For example, if a login times out, the system needs to force the modal window to close to prevent user information leaks. When implementing this functionality, I occasionally encounter the following error:

<img width="965" height="86" alt="Image" src="https://github.com/user-attachments/assets/2d3606e1-1776-4718-98b1-2b390aae1a1e" />

To reproduce the error, you can add the following code to the modal window

<img width="462" height="283" alt="Image" src="https://github.com/user-attachments/assets/74b863cf-95fc-4be2-8ff3-ff0666c1b689" />

### What type of frontend framework are you seeing the problem on?

Angular (Standalone)

### On which version of the frontend framework are you experiencing the issue?

@angular/core@20.0.0,

### Which version of iX do you use?

v5.0.0

### Code to produce this issue.

```shell
https://stackblitz.com/edit/dsaodbqk?file=src%2Fmodal-by-instance-content.ts
```

chencc1988:
Additional note: When a window is minimized, if a Redux event first triggers the window to open and then triggers it to hide—regardless of the time interval between opening and closing—the above error will always occur when the window is restored.

PR #2621, opened 2026-06-23
fix(angular/modal): Fix programmatic modal closing

dmytro-halimov:
GitHub Issue Number: #2595 
JIRA Ref:  #IX-4365

## 💡 What is the new behavior?

[Angular package]: Fixes modal closing race condition where `close()` or `dismiss()` called before modal element initialization would silently fail. Pending actions are now queued and executed once the modal element is set.

## 🏁 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a race condition where modal `close()`/`dismiss()` could be missed if called before the modal was fully initialized.
  * Modal actions are now queued and automatically executed once the modal becomes ready.
  * If both `close` and `dismiss` are requested early, only the first pending action is applied to keep behavior consistent.
* **Tests**
  * Expanded modal service coverage to validate immediate vs deferred close/dismiss behavior and the “first action wins” scenario.

### DIFF
--- a/.changeset/fix-modal-closing-race.md
+++ b/.changeset/fix-modal-closing-race.md
@@ -1,0 +1,7 @@
+---
+'@siemens/ix-angular': patch
+---
+
+Fix modal closing race condition where `close()` or `dismiss()` called before modal element initialization would silently fail. Pending actions are now queued and executed once the modal element is set.
+
+Fixes #2595
--- a/packages/angular/common/src/providers/modal/modal-ref.ts
+++ b/packages/angular/common/src/providers/modal/modal-ref.ts
@@ -8,8 +8,19 @@
  */
 import { closeModal, dismissModal } from '@siemens/ix';
 
+type PendingModalAction<TReason> =
+  | {
+      type: 'close';
+      reason: TReason;
+    }
+  | {
+      type: 'dismiss';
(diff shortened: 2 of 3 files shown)


danielleroux:
@dmytro-halimov Just an idea my side would it make sense to have some QoL feature like a `data-ix-modal-close` and a `data-ix-modal-dismiss` which can be placed of an any element inside the `ix-modal` this will then perform the method call?

danielleroux (changes requested):
- Changeset missing
- PR Title is misleading change is not core related only angular related
- Adapt description to make clear that only the angular service is affected

nuke-ellington (approved):

dmytro-halimov:
> @dmytro-halimov Just an idea my side would it make sense to have some QoL feature like a `data-ix-modal-close` and a `data-ix-modal-dismiss` which can be placed of an any element inside the `ix-modal` this will then perform the method call?

yeah I think this would make sense for core package. But if I get it correctly, these data attributes would only handle user clicks after initial render happens, so it wouldn't cover these programmatic race conditions, the queue in Angular package is still needed.
I can maybe create a ticket for it or do you prefer to do it in this scope?

danielleroux (approved):

PR #2666, opened 2026-07-21
fix(core): stabilize flaky component tests

danielleroux:
## 🆕 What is the new behavior?

- stabilize flaky component tests

## 🏁 Checklist

A pull request can only be merged if all of these conditions are met (where applicable):

- [ ] 🦮 Accessibility (a11y) features were implemented
- [ ] 🗺️ Internationalization (i18n) - no hard coded strings
- [ ] 📲 Responsiveness - components handle viewport changes and content overflow gracefully
- [ ] 📕 Add or update a Storybook story
- [ ] 📄 Documentation was reviewed/updated [siemens/ix-docs](https://github.com/siemens/ix-docs)
- [x] 🧪 Unit tests were added/updated and pass (`pnpm test`)
- [ ] 📸 Visual regression tests were added/updated and pass ([Guide](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md#visual-regression-testing))
- [ ] 🧐 Static code analysis passes (`pnpm lint`)
- [x] 🏗️ Successful compilation (`pnpm build`, changes pushed)

## Related

- EIX-111

## Bug Fixes
- Menus now remain expanded after item selection when pinned.
- Dropdowns open more reliably after rendering and when triggers are added dynamically.
- Date picker keyboard navigation keeps focus on the expected day and prevents unintended scrolling.
- Tabs and About panels activate reliably while content initializes.

## Accessibility
- Improved ARIA forwarding and normalization, including labels and toggle state attributes.

## Tests
- Expanded regression and accessibility coverage for menus, dropdowns, tabs, and form controls.

PR #2677, opened 2026-07-28
fix(core/menu-category): Fix unit test race condition

lzeiml:
Currently this unit test sometimes fails due to a race condition. categoryButton.focus() was called without ensuring tabindex initialization for the categoryButton was complete.
This change prevents this by waiting for aria-posinset to be intialized, signaling that roving-tabindex intialization is complete.

## Summary by CodeRabbit

* **Tests**
  * Added a regression assertion verifying the category button’s accessible position before keyboard navigation checks.


lzeiml:
already addressed with https://github.com/siemens/ix/pull/2666

Issue #1402, opened 2024-07-24
Missing active state for dropdown


katjaKoppensteiner:
### Prerequisites

- [X] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [X] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

The dropdown button is missing it's active state
![image](https://github.com/user-attachments/assets/79ef7dec-a9c4-4272-8870-7ef0f6c790b8)

This leads to the fact that it seems that the dropdown is flying around instead of sticked/belonging to the button.

![image](https://github.com/user-attachments/assets/b63e70a6-6cc5-4eda-a027-a8be140bd96e)

### What type of frontend framework are you seeing the problem on?

Angular

### Which version of iX do you use?

2.3.0

### Code to produce this issue.

```shell
happening on website as well
```

katjaKoppensteiner:
@kathrinschalber This improvement/issue still exists or? 😊

kathrinschalber:
bump

katjaKoppensteiner:
bump

katjaKoppensteiner:
bump

katjaKoppensteiner:
bump

PR #2684, opened 2026-07-30
fix(core): show active state on open dropdown button EIX-84

alexkaduk:

While a dropdown menu is open, `ix-dropdown-button` does not keep a lasting active visual state on the trigger. The trigger can look idle even though `aria-expanded` is `true`, so the open menu can appear disconnected from its control.

GitHub Issue Number: #1402  
Jira: EIX-84

- While the dropdown is open (`aria-expanded="true"`), the inner button’s theme CSS variables are remapped to their `--active` counterparts (including hover → active) so the trigger keeps the Figma Active look for the whole Expanded state.
- Label and icon-only triggers are covered; nested usages that compose `ix-dropdown-button` (e.g. breadcrumb truncate, date picker month/year, split button, tabs overflow) pick up the same behavior.
- Component tests assert the open trigger background matches the Active button token.

## 🏁 Checklist

- [ ] 🦮 Accessibility (a11y) features were implemented *N/A — visual Expanded state only; `aria-expanded` already present*
- [ ] 🗺️ Internationalization (i18n) - no hard coded strings *N/A*
- [ ] 📲 Responsiveness - components handle viewport changes and content overflow gracefully *N/A*
- [ ] 📕 Add or update a Storybook story *N/A — existing previews sufficient*
- [ ] 📄 Documentation was reviewed/updated [siemens/ix-docs](https://github.com/siemens/ix-docs) *N/A — docs site will reflect behavior via existing Usage examples; no API docs change*
- [x] 🧪 Unit tests were added/updated and pass (`pnpm test`) *dropdown-button CT updated; `pnpm run test.ct -- --grep dropdown-button` passed*
- [ ] 📸 Visual regression tests were added/updated and pass ([Guide](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md#visual-regression-testing)) *open-state snapshot likely needs update after intentional UI change*

## Summary by CodeRabbit

* **New Features**
  * Dropdown buttons now display an active visual state while their menu is open.
  * Expanded and collapsed states are shown consistently across supported button variants.
  * Button styling remains consistent across primary, secondary, tertiary, subtle, and danger variants.

* **Bug Fixes**
  * Improved visual connection between an open dropdown menu and its trigger button.

* **Tests**
  * Added visual coverage for expanded and collapsed dropdown-button states.


JonaMaiss (approved):

JonaMaiss (dismissed):

packages/core/src/components/dropdown-button/dropdown-button.ct.ts
```diff
@@ -103,6 +104,92 @@ regressionTest(
...
+  `);
+
+    const dropdownButton = page.locator('ix-dropdown-button');
+    const trigger = dropdownButton.locator('ix-button');
+
+    const getTriggerBackground = () =>
+      trigger.evaluate((el) => {
+        const button = el.shadowRoot?.querySelector('button');
+        return button ? getComputedStyle(button).backgroundColor : '';
+      });
```
lzeiml: can use playwright locator here instead of querying shadowroot
```suggestion
    const getTriggerBackground = async () =>
      trigger
        .locator('button')
        .evaluate((el) => getComputedStyle(el).backgroundColor);
```

packages/core/src/components/dropdown-button/dropdown-button.ct.ts
```diff
@@ -103,6 +104,92 @@ regressionTest(
...
+
+    const closedBackground = await getTriggerBackground();
+
+    await dropdownButton.click();
```
lzeiml: dropdownButton should be checked for hydration before interaction
```suggestion
    await expect(dropdownButton).toHaveClass(/\bhydrated\b/);
    await dropdownButton.click();
```

lzeiml (changes requested):

packages/core/src/components/dropdown-button/dropdown-button.vars.scss
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * SPDX-License-Identifier: MIT
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ */
+
+// Component token aliases. Keep --theme-* references in this file only.
+
+$dropdown-button-variants: primary, secondary, tertiary, subtle-primary,
```
nuke-ellington: These are duplicated with `all-btn-variants` mixin in [button-mixin.scss](url). Can we share them between both?

packages/core/src/components/dropdown-button/dropdown-button.ct.ts
```diff
@@ -103,6 +104,92 @@ regressionTest(
...
+    await expect(dropdownButton).toHaveAttribute('aria-expanded', 'true');
+    await expect(dropdownButton.locator('ix-dropdown')).toBeVisible();
+
+    const openBackground = await getTriggerBackground();
+    expect(openBackground).not.toBe(closedBackground);
+
+    const activeBackground = await dropdownButton.evaluate((el) => {
```
nuke-ellington: This seems tedious. Should this not be apparent in the VRT snapshot? Which did not change BTW 🤔

nuke-ellington (changes requested):

packages/core/src/components/dropdown-button/dropdown-button.ct.ts
alexkaduk: adjusted. thanks

packages/core/src/components/dropdown-button/dropdown-button.ct.ts
alexkaduk: adjusted. thanks

packages/core/src/components/dropdown-button/dropdown-button.vars.scss
alexkaduk: shared that list as $btn-variants in button-mixin.scss and reuse it from dropdown-button.vars.scss
thanks

packages/core/src/components/dropdown-button/dropdown-button.ct.ts
alexkaduk: later removed it, as VRT is used and it presents results much more cleaner.

packages/core/src/components/dropdown-button/dropdown-button.ct.ts
alexkaduk: removed. and add VRT. thanks!

lzeiml (dismissed):

packages/core/src/components/dropdown-button/dropdown-button.vars.scss
lzeiml: might be worth to have the button-variants list in a separate file that can be imported by both button-mixin and dropdown-button.vars to avoid duplication of the list

alexkaduk: done, thanks

nuke-ellington (approved):

Issue #2664, opened 2026-07-21
<ix-application-menu> Scroll doesn't work correctly


carolgad19:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

When I open the side menu since I have a subfolder that has many entries, and the scroll does not work on it, it only works if I am in one of the subfolders' pages. This is a big constraint in our website. Hope it can be fixed swiftly.

<img width="382" height="1969" alt="Image" src="https://github.com/user-attachments/assets/f9c04795-8b3c-44d1-83d1-dd9c533736f4" />

In a subfolders' page

<img width="381" height="1977" alt="Image" src="https://github.com/user-attachments/assets/5b38cd7d-c6ee-43cd-80d8-fea4e42bcdf7" />

Thank you for your work!

### What type of frontend framework are you seeing the problem on?

React

### On which version of the frontend framework are you experiencing the issue?

v18.3.1

### Which version of iX do you use?

v4.2.0

### Code to produce this issue.

```shell
export default function MainMenu(props: Props) {
  const parsersQuery = trpc.parsers.list.useQuery(
    { retrieveCount: true },
    { staleTime: 2 * TimeUtils.miliseconds.MINUTE }
  );

  const parserCount = React.useMemo(
    () => parsersQuery.data?.reduce((sum, family) => sum + (family.count ?? 0), 0),
    [parsersQuery.data]
  );

  const user = useSessionUser();

  return (
    <IxMenu>
      <IxMenuAvatar
        top={user.email}
        bottom={`${Utils.formatUserName(user)} (${user.department})`}
        image={`https://ui-avatars.com/api/?name=${user.givenName.charAt(0)}+${user.familyName.charAt(0)}`}
      />

      <NavLink to="/">
        {({ isActive }: { isActive: boolean }) => (
          <IxMenuItem home icon={iconHome} active={isActive}>
            Home
          </IxMenuItem>
        )}
      </NavLink>
      <IxMenuItem icon={iconPiechart} style={{ display: "none" }}>
        Dashboard
      </IxMenuItem>
      <NavLink to="/cves">
        {({ isActive }: { isActive: boolean }) => (
          <IxMenuItem icon={iconMonitor} active={isActive}>
            CVE Search
          </IxMenuItem>
        )}
      </NavLink>
[...]

PR #2706, opened 2026-08-09
fix(core): prevent menu-category expand clipping tall items EIX-117

benjgil:

If an application sub-menu contains enough items to create a scroll, the scroll is only visible when the user is on a page that is in that sub-menu. If the user is on a page that is NOT inside the long sub-menu, and then hovers on the main nav menu item that contains the sub-menu, then the sub-menu scroll is not visible and items in the sub-menu are not clickable

GitHub Issue Number: #2664 

Fix the bug by setting showDropdown to true before the dropdown children and scroll are calculated

## 🏁 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Menu category dropdowns now open reliably and display all items correctly.
  * Long dropdown menus are scrollable, including when no category item is active, keeping the final item accessible.
  * Selecting an item from a long dropdown now closes the menu as expected.
  * Expanded menus no longer clip oversized content or leave unnecessary size constraints behind.

* **Tests**
  * Added regression coverage for large dropdowns, scrolling, item visibility, and selection behavior.


packages/core/src/components/menu-category/test/menu-category.ct.ts
```diff
@@ -547,6 +547,55 @@ regressionTest(
   }
 );
 
+regressionTest(
+  'should allow scrolling in dropdown with many items when no category item is active',
```
alexkaduk: The test name says scrolling, but we only assert scrollHeight > clientHeight and not.toBeHidden(). That does not prove the last item is reachable. Please follow the same pattern as dropdown.ct.ts (scrollIntoView + visible / click).

packages/core/src/components/menu-category/menu-category.tsx
```diff
@@ -173,6 +173,8 @@ export class MenuCategory
     }
     this.closeOtherCategories.emit(this.categoryId);
 
+    this.showDropdown = true;
```
alexkaduk: Why set showDropdown = true before present()? 
The real need is to clear hidden on slotted items before overflow is measured (onShowItemsChange). Setting show early can open the dropdown even if willPresent() / showChange was canceled. Can we unhide the items directly and keep showDropdown synced from showChanged?

alexkaduk:
Please rename the PR title to conventional format: fix(core/menu-category): … (see .github/instructions/code-review.instructions.md).

alexkaduk:
Consider to add Fixes #2664 to the changeset summary so the issue closes on release.

packages/core/src/components/menu-category/menu-category.tsx
benjgil: Good point, though on the flip side, it's not good to manually set hidden for all children without it being clear where or why that's happening and bypassing existing mechanisms for doing that.

I prefer implementing a separate flag, willShowItems, that signifies the intermediate period for preparing to show menu items, which is what this is, and connecting that to the existing logic for hiding the sub-menu items

packages/core/src/components/menu-category/test/menu-category.ct.ts
benjgil: the issue wasnt reachability but that the scroll itself didnt exist, which is being tested here - and because the actual scroll functionality itself is already tested in dropdown.ct.ts there's no real need to test the dropdown functionality here again. 
But i've added it anyway now as it doesnt hurt to make sure we're using the dropdown correctly end to end.

.changeset/menu-category-dropdown-scroll.md
```diff
@@ -0,0 +1,5 @@
+---
+'@siemens/ix': patch
+---
+
+fix(core/menu-category): ensure sub-menu items are visible before dropdown presents so scrolling works correctly when no category item is active
```
alexkaduk: Fixes #2664 is missing. pls add

benjgil: done

alexkaduk (dismissed):

packages/core/src/components/menu-category/menu-category.tsx
```diff
@@ -430,25 +426,23 @@ export class MenuCategory
...
-          this.clearMenuItemStyles();
+          this.clearMenuItemsContainertyles();
         }
         this.showItems = this.isCategoryItemListVisible();
       }
     );
   }
 
-  clearMenuItemStyles() {
+  clearMenuItemsContainertyles() {
```
nuke-ellington:
```suggestion
  clearMenuItemsContainerStyles() {
```

.changeset/menu-category-dropdown-scroll.md
```diff
@@ -0,0 +1,7 @@
...
```
nuke-ellington: Changeset is supposed to be Human readable. Commit message format is not appropriate here :)

packages/core/src/components/menu-category/test/menu-category.ct.ts
```diff
@@ -547,6 +547,102 @@ regressionTest(
...
+    const menuCategory = page.locator('ix-menu-category').first();
+
+    // Override item heights to be larger than the hardcoded 40px used by
+    // getNestedItemsHeight(), so the animation's inline max-height
+    // underestimates the actual content and overflow:hidden clips items.
+    await menuCategory.evaluate((cat: HTMLIxMenuCategoryElement) => {
+      cat.querySelectorAll(':scope ix-menu-item').forEach((item) => {
+        (item as HTMLElement).style.setProperty(
+          '--ix-menu-item-height',
+          '80px'
```
nuke-ellington: This seems a bit artificial. Can this test be more closer to the actual use case?

nuke-ellington (changes requested):

packages/core/src/components/menu-category/test/menu-category.ct.ts
benjgil: the use case is varied, according to screen size, browser window height, number of elements in the sub-menu etc. What we need to test is that the menu height is calculated properly and scroll isnt hidden as per the bug, and this is the most fool-proof way to do that

.changeset/menu-category-dropdown-scroll.md
benjgil: done

packages/core/src/components/menu-category/test/menu-category.ct.ts
benjgil: agreed - i misread this that it was setting the height of the menu container, not each menu item. Thanks for catching.

packages/core/src/components/menu-category/menu-category.tsx
benjgil: In the end implemented it entirely differently after seeing a better alternative while making the above intended change

alexkaduk (approved):

nuke-ellington (approved):

Issue #2689, opened 2026-08-03
ix-event-list: triggerFadeOut animates a null target, logging "No target found" on every child mutation


wbarcik-etm:
# Bug: `ix-event-list` logs "No target found" from anime.js on every child mutation

## Describe the bug

`ix-event-list` animates a `null` target on every DOM mutation of its children, logging this anime.js warning to the console:

```
No target found. Make sure the element you're trying to animate is accessible before creating your animation.
```

It fires even when `animated` is not set (i.e. `false`, the default). In an app that updates list contents from a WebSocket push, this floods the console.

Two separate defects in `triggerFadeOut()` combine to cause it. Both are present on `main`:
https://github.com/siemens/ix/blob/main/packages/core/src/components/event-list/event-list.tsx

```ts
private triggerFadeOut(): Promise<void> {
  return new Promise((resolve) => {
    if (!this.animated) {
      resolve();                    // (1) missing `return` — execution falls through
    }
    const listElement = this.hostElement.shadowRoot!.querySelector('ul');
    animate(listElement!, {         // (2) listElement is always null
      opacity: [{ opacity: 1, easing: 'easeInSine' }, { opacity: 0 }],
      duration: EventList.fadeOutDuration,
      onComplete: () => { resolve(); },
    });
  });
}
```

**(1) Missing `return` after `resolve()`.** Resolving a promise does not stop execution, so the `animate()` call runs regardless of `animated`. The intent was clearly an early exit — `triggerFadeIn()` directly above does it correctly with `if (!this.animated) { return; }`.

**(2) `querySelector('ul')` never matches.** `render()` emits a `<div role="list">`, not a `<ul>`:

```tsx
render() {
  return (
    <Host class={{ ... }}>
      <div role="list">
        <slot></slot>
      </div>
    </Host>
  );
}
```

So `listElement` is always `null`, and the non-null assertion `animate(listElement!, ...)` hides it from the type checker. This means the fade-out animation **never works at all**, even for consumers who do set `animated` — the warning is just the visible symptom.

The trigger path is `componentDidLoad()` registering a subtree observer, whose callback calls `triggerFadeOut()` unconditionally:

```ts
this.mutationObserver.observe(this.hostElement, { childList: true, subtree: true });
```

```ts
private onMutation(mutationRecords: Array<MutationRecord>) {
  this.triggerFadeOut().then(() => { ... });
}
```

So any change to slotted content — adding, removing or re-rendering an `ix-event-list-item` — logs the warning.

[...]

PR #2710, opened 2026-08-11
fix(core): prevent event-list anime.js no-target warning EIX-177

benjgil:

`ix-event-list` triggers 'no element found' warnings due to incorrect DOM querying, and non-returned promise resolve.

GitHub Issue Number: #2689 

- returning when animated is `false` ensures the animation logic only runs when the user configures it to
-  Corrected DOM query ensures that the fadeOut works properly when animation is `true`

## 🏁 Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed event list transitions when animations are disabled.
  * Prevented console warnings when dynamically adding event list items, with or without animations enabled.
  * Improved animation targeting for the rendered event list container.
  * Improved event list interaction reliability while content is loading.

* **Tests**
  * Added regression coverage for dynamically appended event list items.
  * Added accessibility checks confirming the event list has no detected violations.


alexkaduk:
changeset-bot correctly flags no changeset. This is a consumer-visible bug fix.

per [CONTRIBUTING.md](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md) (Submit Pull Request §4) user-facing bug fixes need a .changeset/ entry.

changeset create skill might be used.

packages/core/src/components/event-list/test/event-list.ct.ts
```diff
@@ -19,6 +19,66 @@ import 'jest';
...
+    `);
+
+    await page.evaluate(() => {
+      const list = document.querySelector('ix-event-list');
+      const item = document.createElement('ix-event-list-item');
+      item.textContent = 'Item 2';
+      list!.appendChild(item);
+    });
+
+    await page.waitForTimeout(200);
```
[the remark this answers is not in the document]
alexkaduk: comment is valid. per .github/instructions/component-tests.instructions.md and AGENTS.md: Avoid arbitrary timeouts (waitForTimeout). pls check and fix.

packages/core/src/components/event-list/event-list.tsx
```diff
@@ -120,10 +120,11 @@ export class EventList {
   private triggerFadeOut(): Promise<void> {
     return new Promise((resolve) => {
       if (!this.animated) {
-        resolve();
+        return resolve();
       }
 
-      const listElement = this.hostElement.shadowRoot!.querySelector('ul');
+      const listElement =
+        this.hostElement.shadowRoot!.querySelector('div[role="list"]');
```
alexkaduk: Can we also guard a missing list container and drop listElement! / prefer optional chaining on shadowRoot, as suggested in the issue? The ! is what hid the original ul mismatch.

packages/core/src/components/event-list/test/event-list.ct.ts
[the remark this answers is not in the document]
alexkaduk: @benjgil pls check all tests for the same waitForTimeout

packages/core/src/components/event-list/test/event-list.ct.ts
```diff
@@ -19,6 +19,79 @@ import 'jest';
...
+      const list = document.querySelector('ix-event-list');
+      const item = document.createElement('ix-event-list-item');
+      item.textContent = 'Item 2';
+      list!.appendChild(item);
+    });
+
+    await page.waitForTimeout(300);
+
+    const animeWarnings = warnings.filter((w) => w.includes('No target found'));
+    expect(animeWarnings).toHaveLength(0);
```
alexkaduk: Both tests couple to the anime.js “No target found” string — if that message changes they stay green and still burn waitForTimeout.
Prefer an assertion we control over the third-party console text.

benjgil: Unfortunately the whole issue that we're checking is with an error from the third part library, so we dont control anything here. 

Any suggestion for how to get round that?

packages/core/src/components/event-list/test/event-list.ct.ts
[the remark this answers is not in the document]
benjgil: @alexkaduk 
Page.waitfortimeout is used across all of the component tests in the entire project. 
If we want to remove that pattern then lets create a new task for that, it's not in the scope of this task/PR

I'll remove it for the new tests in this PR

packages/core/src/components/event-list/event-list.tsx
benjgil: Agreed - there's no reason to force it here either way

packages/core/src/components/event-list/test/event-list.ct.ts
alexkaduk: What we do control is our animate(...) call site: when animated is false it must not run, and when true it must target div[role="list"] (not null/ul). 
Prefer asserting that (e.g. mocked animate in a spec) over matching the third-party console string.

benjgil: ok, happy to go with that

packages/core/src/components/event-list/test/event-list.ct.ts
```diff
@@ -19,6 +19,83 @@ import 'jest';
...
+    await page.evaluate(() => {
+      const list = document.querySelector('ix-event-list');
+      const item = document.createElement('ix-event-list-item');
+      item.textContent = 'Item 2';
+      list!.appendChild(item);
+    });
+
+    await expect(page.locator('ix-event-list-item').nth(1)).toHaveClass(
+      /hydrated/
+    );
```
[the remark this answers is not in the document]
benjgil: anyItemHasAnimatedStyle now checks that triggerFadeOut was (or wasn't) called

packages/core/src/components/event-list/test/event-list.ct.ts
```diff
@@ -17,7 +17,82 @@
...
+
+const anyItemHasAnimatedStyle = (page: Page) =>
+  page.evaluate(() => {
+    const items = document.querySelectorAll('ix-event-list-item');
+    return Array.from(items).some(
+      (item) => (item as HTMLElement).style.opacity !== ''
+    );
+  });
+
+regressionTest(
```
alexkaduk: Thanks for the try to improve the the tests. At the same time, let me remove the newly added animation regression tests and left the existing CT coverage as on main.

Reasoning:

- Those tests still pass against main where the anime.js "No target found" warning is present, so they don’t actually catch the bug.
- The failure mode is a third-party console warning from anime.js. Asserting that message (or trying to reverse-engineer anime’s DOM/style side effects) is brittle and doesn’t give a reliable regression signal.
- The PR keeps the small core fix in triggerFadeOut. 

I’ve verified the warning is gone.

alexkaduk (approved):

lzeiml (approved):

.changeset/event-list-no-target-warning.md
```diff
@@ -0,0 +1,7 @@
+---
+'@siemens/ix': patch
+---
+
+Fix `ix-event-list` so child mutations no longer call anime.js with a missing target, which stopped the console `"No target found"` warning (including when `animated` is unset).
```
benjgil, line 5: this has already been merged, but the text is missing a key part of the fix - the bug included that the animated attribute wasnt actually doing anything because it was not targeting the correct part of the shadow dom. 

The original changeset description included that

PR #2724, opened 2026-08-18
feat(core/list): Implement generic list component

lzeiml:
EIX-65
Docs PR: https://github.com/siemens/ix-docs/pull/279

Add ix-list, ix-list-item and ix-list-separator.
Implements the following functionality:
- Keyboard navigable list with standard/custom content
- Checkbox selection
- Variants
- Dividers/spacing
- Action slot (persisent/on-hover)
- Drag and Drop (dynamic/separator)

## 🏁 Checklist

## Summary by CodeRabbit

* **New Features**
  * Added versatile list, list-item, and separator components.
  * Supports selection, custom content, descriptions, dividers, spacing, disabled states, visual variants, and hover actions.
  * Added pointer- and keyboard-based drag-and-drop reordering.
  * Added keyboard navigation and accessibility-focused interactions.
  * Added framework examples and documentation previews for common list configurations.

* **Tests**
  * Added accessibility, interaction, reordering, and visual regression coverage.


alexkaduk:
Thanks for the CodeRabbit pass. Please treat these as must-fix before merge (they are real, not nits). Details can be found on the CodeRabbit comments. Items worth checking:

1. **Keyboard activation on draggable lists** — Enter/Space on the primary action currently start reorder and swallow `itemClick`. Reorder should stay on the gripper; primary action should still activate (add a ct for Enter → `itemClick` with `draggable`).
2. **`list-item.ct.ts` is out of date** — `status` / `ix-pill` / hidden `.label`, and selected color vs `--theme-color-ghost--selected`, do not match the current component.
3. **`tooltipText`** — public API with the tooltip commented out. Implement it or remove the prop (and the dead JSX) before v6.
4. **`aria-labelledby`** — the explicit `aria-labelledby={undefined}` overwrites an inherited `aria-labelledby` from `inheritAriaAttributes`. Preserve the inherited value.
5. **Reorder robustness**
   - Guard `pointercancel` the same way as `pointerup` (`dragMode === 'pointer'` + matching `pointerId`), so touch cancel cannot abort a keyboard reorder.
   - Cancel/finish keyboard reorder on focus leaving the list (Tab out currently leaves `dragging` stuck).
   - `has-divider` still forces placeholder `height: 0.25rem !important` in **dynamic** drag (not only separator). Please scope that rule; items jump during drag.
6. **MutationObserver** — start `observe` in `connectedCallback` so a disconnected/reconnected list keeps syncing defaults and tabindex.
7. **Release** — this is targeted at **v6**. Keep `@since 6.0.0`; please change the changeset from `minor` to `major` (do **not** follow the CodeRabbit note to retag as 5.3.0).
8. **Framework smoke** — generated `generic-list-*.spec.ts` files are not imported in `working.spec.ts` / `working-with-axe.spec.ts`.

Might make sense to apply SPDX on new SCSS, HTML titles (“Event list example”), Storybook separator/`actionOnHover` stories, localizable gripper/select labels, preview `addIcons` placement.

packages/core/src/components/list-item/list-item.tsx
```diff
@@ -0,0 +1,301 @@
...
+  'ix-button',
+  'ix-checkbox',
+  'ix-dropdown-button',
+  'ix-icon-button',
+  'ix-split-button',
+  'ix-toggle',
+].join(',');
+
+/**
+ * @slot default - Non-interactive custom content replacing the standard icon, label, description, and status layout.
```
alexkaduk: This says default-slot content replaces icon, label, description, and status. There is no status API anymore, and standard content still renders whenever icon / label / description are set. Please adjust the JSDoc to match the current contract

testing/visual-testing/tests/list/basic/index.html
```diff
@@ -0,0 +1,53 @@
...
+        max-height: 18rem;
+      }
+    </style>
+  </head>
+  <body>
+    <ix-list aria-label="Projects" draggable has-divider item-gap="4">
+      <ix-list-item
+        label="Factory overview"
+        description="Updated 5 minutes ago"
+        status="Online"
```
alexkaduk: This fixture still sets status="Online" / status="Warning" and slot="additional-actions", which ix-list-item does not support. Visual snapshots will not show the intended UI. Please update the fixture to current slots/props before trusting the screenshots.

packages/react-test-app/src/preview-examples/generic-list-additional-options.tsx
```diff
@@ -0,0 +1,79 @@
...
+  iconEditDocument,
+  iconProject,
+  iconTrashcan,
+} from '@siemens/ix-icons/icons';
+
+export default function GenericListAdditionalOptions() {
+  addIcons({ iconEditDocument, iconProject, iconTrashcan });
+
+  return (
+    <IxList checkbox>
```
alexkaduk, line 35: Selection is controlled (selectedChange does not write selected). Storybook Selection wires the event; this preview (and the HTML/Angular/Vue copies) do not, so the checkboxes snap back. Please handle selectedChange in all five test apps, or show a static selected item so the example is honest.

packages/storybook-docs/src/stories/list.stories.ts
```diff
@@ -0,0 +1,123 @@
...
+
+  return container;
+};
+
+const meta = {
+  title: 'Example/List',
+  tags: [],
+  render: (args) => renderList(args),
+  argTypes: makeArgTypes<Partial<ArgTypes<Element>>>('ix-list', {}),
+  parameters: {},
```
alexkaduk: parameters is empty and there is no list/accessibility story with a11y: { test: 'error' }. Global preview a11y is still test: 'todo'. CodeRabbit already asked for separator / actionOnHover stories; a11y stories can be a follow-up.

packages/vue-test-app/src/preview-examples/generic-list-variants.vue
```diff
@@ -0,0 +1,40 @@
...
+
+<template>
+  <div style="display: flex; flex-direction: column; gap: 24px">
+    <IxList has-divider item-gap="0" variant="ghost">
+      <IxListItem icon="project" label="Item 1"></IxListItem>
+      <IxListItem icon="project" label="Item 2"></IxListItem>
+      <IxListItem
+        icon="project"
+        label="Item 3"
+        has-divider="false"
```
alexkaduk, line 26: React correctly uses hasDivider={false}. Here (and in both Angular generic-list-variants.html files) this is the string "false". Elsewhere in Vue we bind booleans (:range="false"). Please consider to use :has-divider="false" / [hasDivider]="false" so item 3 actually turns the divider off.

packages/html-test-app/src/preview-examples/generic-list-basic.html
```diff
@@ -0,0 +1,32 @@
...
+This source code is licensed under the MIT license found in the
+LICENSE file in the root directory of this source tree.
+-->
+
+<!DOCTYPE html>
+<html lang="en" data-ix-theme="classic" data-ix-color-schema="system">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Event list example</title>
```
alexkaduk: Same title is on all generic-list-*.html previews. Might make sense to rename to “List example” (or similar). CSS class names in custom-content (event-list-*) have the same leftover if you want to clean those up too

packages/core/src/components/list-item/list-item.scss
```diff
@@ -0,0 +1,268 @@
...
+  @include fonts.text-default;
+}
+
+.description {
+  @include fonts.text-s;
+
+  color: var(--theme-color-soft-text);
+}
+
+.status,
```
alexkaduk: .status is grouped with .action but there is no status markup anymore.
is it needed?

alexkaduk:
One more item worth checking: please use the **v6 design tokens** (SI Theme 6 / `--si-sys-*`) instead of legacy `--theme-*` and `legacy/mixins/fonts`. This component is new for v6, so it should ship on the new system rather than needing a follow-up migration.

Guide: `BREAKING_CHANGES/v6.md` (map by **purpose**, not by prefix). Sass can use:

```scss
@use '@siemens/ix/scss/tokens/system' as sys;
```

Places that still use the old tokens:

- `packages/core/src/components/list-item/list-item.scss` — `--theme-color-std-text`, `--theme-color-ghost*`, `--theme-color-component-1*`, `--theme-color-soft-bdr`, `--theme-color-focus-bdr`, `--theme-color-soft-text`, `--theme-color-weak-text`, plus `@use 'legacy/mixins/fonts'` (`text-default` / `text-s`)
- `packages/core/src/components/list/list.scss` — `--theme-shadow-1`, `--theme-color-std-bdr`, `--theme-color-contrast-bdr`
- `packages/core/src/components/list-item-separator/list-item-separator.scss` — `--theme-color-soft-bdr`
- Preview CSS (`generic-list-custom-content.css` in the test apps) — `--theme-color-soft-text`

Expected direction (from the v6 mapping; please confirm against design):

- `--theme-color-std-text` → `--si-sys-text-primary`
- `--theme-color-soft-text` → `--si-sys-text-secondary`
- `--theme-color-weak-text` → `--si-sys-text-disabled`
- `--theme-color-ghost` → `transparent`
- `--theme-color-ghost--hover` / `--theme-color-component-1--hover` → `--si-sys-background-hover`
- `--theme-color-ghost--active` / `--theme-color-ghost--selected` / `--theme-color-component-1--active` → `--si-sys-background-selected`
- `--theme-color-component-1` → `--si-sys-background-1`
- `--theme-color-soft-bdr` → `--si-sys-border-3`
- `--theme-color-std-bdr` → `--si-sys-border-2`
- `--theme-color-contrast-bdr` → `--si-sys-border-1`
- `--theme-color-focus-bdr` → `--si-sys-effects-focus`
- `--theme-shadow-1` → `--si-sys-effects-shadow-1` (full `box-shadow` value)
- fonts: `--si-sys-typography-*` (complete `font` shorthand), not the legacy mixins

Please also update the selected-state ct that currently reads `--theme-color-ghost--selected`, so it asserts the token the styles actually use after the migration.

packages/react-test-app/src/preview-examples/generic-list-additional-options.tsx
lzeiml: adjusted to behave like storybook

testing/visual-testing/tests/list/basic/index.html
lzeiml: fixed

packages/core/src/components/list-item/list-item.tsx
lzeiml: fixed

packages/storybook-docs/src/stories/list.stories.ts
lzeiml: fixed a11y test and added the stories

packages/vue-test-app/src/preview-examples/generic-list-variants.vue
lzeiml: Using `:has-divider="false"` / `[hasDivider]="false"` causes `ix-list`'s `applyItemDefaults()` (which checks `item.hasAttribute('has-divider')`) to miss the override and overwrite `hasDivider` back to `true` from the parent `<ix-list has-divider>`. Because boolean `false` does not set a DOM attribute, passing the string `has-divider="false"` is currently required so `item.hasAttribute('has-divider')` evaluates to true and preserves the override.

packages/html-test-app/src/preview-examples/generic-list-basic.html
lzeiml: fixed

packages/core/src/components/list-item/list-item.scss
lzeiml: removed the useless class

packages/core/src/components/list/list.tsx
```diff
@@ -0,0 +1,884 @@
...
+    if (
+      isPrimaryAction &&
+      isActivationKey &&
+      this.draggable &&
+      !dragGripper?.disabled
+    ) {
+      event.preventDefault();
+      this.beginReorder(item, 'keyboard');
+      return;
+    }
```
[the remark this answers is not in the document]
lzeiml: removed as not needed

packages/core/src/components/list/list.tsx
```diff
@@ -0,0 +1,1011 @@
...
+  Watch,
+} from '@stencil/core';
+import type {
+  ListItemActionSlotAlignment,
+  ListItemVariant,
+} from '../list-item/list-item';
+import { createMutationObserver } from '../utils/mutation-observer';
+
+export type ListItemGap = 0 | 4 | 8 | 12;
+export type ListDragBehavior = 'dynamic' | 'separator';
```
danielleroux: Create type file and export in public api, see other components

packages/core/src/components/list/list.tsx
```diff
@@ -0,0 +1,1011 @@
...
+
+export interface ListItemOrderChangeEvent {
+  item: HTMLIxListItemElement;
+  oldIndex: number;
+  newIndex: number;
+}
```
danielleroux: Create type file and export in public api

packages/core/src/components/list/list.tsx
```diff
@@ -0,0 +1,1011 @@
...
+  private dragStartY = 0;
+  private itemsSynchronized = false;
+  private readonly inheritedItemValues = new WeakMap<
+    HTMLIxListItemElement,
+    Partial<
+      Record<
+        InheritedItemProperty,
+        ListItemVariant | ListItemActionSlotAlignment | boolean
+      >
+    >
```
danielleroux, line 184: Why do we need here a Partial Record type?

packages/core/src/components/list/list.tsx
```diff
@@ -0,0 +1,1011 @@
...
+    this.dragMode = mode;
+    this.dragOriginalIndex = this.items.indexOf(item);
+    this.dragOriginalNextSibling = item.nextSibling;
+    this.setItemReorderState(item, true);
+    this.announce(
+      item,
+      `${this.getItemLabel(
+        item
+      )} lifted. Use arrow keys to move, Enter or Space to drop, and Escape to cancel.`
+    );
```
danielleroux, line 520: Just put the comment here but hit every hardcoded announce string, we need to provide a way to make it i18n ready

packages/core/src/components/list/list.tsx
```diff
@@ -0,0 +1,1011 @@
...
+    if (this.dragBehavior === 'dynamic') {
+      placeholder.style.height = `${itemBounds.height}px`;
+    }
+    this.hostElement.insertBefore(placeholder, item);
+    this.dragPlaceholder = placeholder;
+
+    if (this.dragBehavior === 'dynamic') {
+      item.style.setProperty('--ix-list-drag-left', `${itemBounds.left}px`);
+      item.style.setProperty('--ix-list-drag-top', `${itemBounds.top}px`);
+      item.style.setProperty('--ix-list-drag-width', `${itemBounds.width}px`);
```
danielleroux: Would make sense to use here also rem

packages/core/src/components/list/list.tsx
danielleroux, line 1: Slotted components lose native keyboard behaviour:

Image

E.G select Home/End jumps to this end and lose select focus

danielleroux, line 1: Transform of parent element brakes the drag and drop:

```tsx
            <div
              style={{
                transform: 'translate(40px, 20px)',
                width: 'calc(100% - 40px)',
                'margin-bottom': '3rem',
              }}
            >
              <ix-list
                draggable
              >
                <ix-list-item label="1" />
                <ix-list-item label="1" />
                <ix-list-item label="3" />
              </ix-list>
```

danielleroux, line 1: Maybe also related to the translate comment, but list drag and drop does not work in any modal like component:

- ix-modal
- ix-popover
- ix-dropdown

Image

packages/core/src/components/list/list.scss
danielleroux, line 1: sys tokens directly consumed, checkout other components, introduce a list.vars.scss

packages/core/src/components/list-item/list-item.scss
danielleroux, line 1: sys tokens directly consumed, checkout other components, introduce a list-item.vars.scss

packages/core/src/components/list-item-separator/list-item-separator.scss
danielleroux, line 1: sys tokens directly consumed, checkout other components, introduce a list-item-seperator.vars.scss

packages/core/src/components/list-item/list-item.tsx
```diff
@@ -0,0 +1,308 @@
...
+
+  private hasStandardContent() {
+    return !!(this.icon || this.label || this.description);
+  }
+
+  override getIgnoredAriaAttributes(): A11yAttributeName[] {
+    return ['role', 'aria-checked', 'aria-disabled', 'aria-pressed'];
+  }
+
+  override componentDidLoad() {
```
danielleroux, line 177: super call missing

packages/core/src/components/list/list.scss
danielleroux, line 1: Not having any lint / style check for this but, look into other components scss files, normally we but all component (:host) stuff inside the host selector not flat into the file

danielleroux (changes requested):

danielleroux (changes requested):
Storybook for "ix-list-item" is missing

packages/core/src/components/list/list.tsx
danielleroux: Makes sense to have that :)

packages/core/src/components/list/list.tsx
lzeiml: i dont think rem makes sense here, because `getBoundingClientRect()` returns the values in px, and writing them  back in px preserves the same system

PR #2729, opened 2026-08-20
fix(core): keep input field active look while overlay is open EIX-151

alexkaduk:

Opening an overlay on field-style triggers (`ix-date-input`, `ix-time-input`, `ix-datetime-input`, `ix-select`, `ix-category-filter`) focuses the field, then focus often moves into the overlay. Native `:hover` / `:focus` on the field chrome is gone, so the field falls back to Default while the overlay stays open.

GitHub Issue Number: #1402 (EIX-151)

- While the overlay is open, the field host gets class `active` (same overlay-open naming as `ix-dropdown-button`). The **look** is field **Hover** via `@include input.element-input-hover` (`--theme-input--border-color--hover`, `--theme-input--background--hover`), not Focus outline and not button Active tokens. Focus outline is left to real focus; it is not forced or cleared for open overlays.
- Shared Hover/Focus token helpers live in `packages/core/scss/mixins/_input.scss` (`element-input-hover`, `element-input-focus`).
- **Date / time / datetime:** `:host(.active)` in `input-field` reapplies Hover on the native `<input>`. Host `active` tracks `this.show`. Calendar/clock `ix-icon-button` stays Subtle tertiary Default; `aria-expanded` uses `a11yBoolean(this.show)` (added on date/datetime; time already had it).
- **Select:** host `active` tracks `dropdownShow`; Hover applied on `.select` (local SCSS — cannot use `input-field`).
- **Category-filter:** `showDropdown` synced from `ix-dropdown` `showChanged`; host `active` tracks it; Hover applied on `.input-container`.

## 🏁 Checklist

- [ ] 📕 Add or update a Storybook story — *N/A (no new variant)*
- [ ] 📄 Documentation was reviewed/updated [siemens/ix-docs](https://github.com/siemens/ix-docs) — *N/A unless usage pages describe picker/select trigger field states*
- [ ] 🧪 Unit tests were added/updated and pass (`pnpm test`) — *N/A (open look covered by visual regression, not class-name CTs)*

## 👨‍💻 Help & support

Manual check in react-test-app — open the overlay, move the pointer off the field, confirm Hover chrome stays:

- `/preview/date-input`
- `/preview/time-input`
- `/preview/datetime-input`
- `/preview/select`
- `/preview/category-filter`

Optional extras: `/preview/select-editable`, `/preview/select-multiple`, `/preview/category-filter-suggestions`

## Summary by CodeRabbit

* **Bug Fixes**
  * Date, time, datetime, select, and category-filter fields now retain their hover appearance while dropdowns are open.
  * Open picker and dropdown states are communicated more accurately to assistive technologies.
  * Active, editable fields now have more consistent styling, while disabled and read-only states remain unchanged.

* **Tests**
  * Added and updated visual regression coverage for open date, time, datetime, select, and category-filter dropdowns.


packages/core/src/components/input/input.mixins.scss
```diff
@@ -109,6 +109,13 @@ $input-end-container-inset: 0.25rem;
     }
   }
 
+  // Picker fields (date/time/datetime) set host `active` while the overlay is open.
+  :host(.active:not(.disabled):not(.readonly)) {
+    input:not(:disabled):not(:read-only) {
+      @include input.element-input-hover;
+    }
+  }
```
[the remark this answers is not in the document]
alexkaduk: it is expected to have outline

lzeiml (changes requested):
~~According to figma, the focus outline should also stay visible when tabbing into the dropdown, but right now it disappears:
image
-->
image~~

Edit: behaviour is intended according to UX

testing/visual-testing/__screenshots__/tests/date-input/date-input.e2e.ts/date-input-dropdown-open-1-chromium---classic-dark-linux.png
lzeiml, line 1: To verify the fix, we should have a case where focus moves inside the dropdown. As it is right now, it only shows the expected behaviour that is already working on the main branch (with focus still on input).

Same for the other visual tests for the affacted components.

lzeiml (changes requested):

testing/visual-testing/__screenshots__/tests/date-input/date-input.e2e.ts/date-input-dropdown-open-1-chromium---classic-dark-linux.png
alexkaduk: tests are updated. thanks

packages/core/scss/mixins/_input.scss
```diff
@@ -6,33 +6,49 @@
...
  */
+@use '../tokens/system' as sys;
 @use './fonts';
 @use './validation/form-component';
 
 @mixin element-input-focus {
-  outline: 1px solid var(--theme-color-focus-bdr);
-  outline-offset: var(--theme-input--focus--outline-offset);
-  border-color: var(--theme-input--border-color--focus) !important;
+  outline: 1px solid #{sys.$si-sys-effects-focus};
```
lzeiml: instead of using the system variable directly here, we should add it as a variable in input.vars.scss

packages/core/scss/mixins/_input.scss
```diff
@@ -6,33 +6,49 @@
...
 @mixin element-input-hover {
-  border-color: var(--theme-input--border-color--hover) !important;
-  background-color: var(--theme-input--background--hover);
+  border-color: var(
+    --ix-input--border-color--hover,
+    #{sys.$si-sys-border-1}
+  ) !important;
+  background-color: var(
+    --ix-input--background--hover,
+    #{sys.$si-sys-background-4}
```
lzeiml: same for these 3 sys variables, they are already mapped in input.vars.scss, maybe those can be reused directly

lzeiml (changes requested):

packages/core/scss/mixins/_input.scss
alexkaduk: updated, thanks

packages/core/scss/mixins/_input.scss
alexkaduk: updated, thanks

lzeiml (approved):

PR #2735, opened 2026-08-20
fix(core): show active state on open group and avatar menus EIX-151

alexkaduk:

Opening the group kebab (`ix-group-context-menu`) or the application-header `ix-avatar` menu applies native `:hover` / `:active` only while the pointer stays on the trigger. Moving into the overlay (or pressing Escape without a lock) drops the trigger back to Default, so it no longer matches Figma **Active** while the dropdown is open.

Header avatar is not an `ix-button`. Putting `active` only on the inner native button did not keep the tertiary Active fill once the pointer left the glyph (scrollbar visual: hover last item).

GitHub Issue Number: #1402 (EIX-151)

- Overlay still owns *when* the menu is open; the button owns *how* it looks. Host class `active` (not a public `active` prop, not `selected`, not a global `aria-expanded` style).
- `ix-group-context-menu` observes the slotted `ix-dropdown` `showChanged` event and sets `active` plus `aria-expanded` / `aria-haspopup` on the kebab `ix-icon-button`. `showContextMenu` remains “kebab visible” (slot has content); `dropdownShow` is “overlay open”.
- Header `ix-avatar` sets `active` on the host (`:host(.avatar-button.active)` locks tertiary Active via `btn-active-look`) and on the inner `BaseButton` (`extraClasses`). `aria-expanded` was already tied to `dropdownShow`.

## 🏁 Checklist

- [ ] 📕 Add or update a Storybook story — *N/A (no new variant)*
- [ ] 📄 Documentation was reviewed/updated [siemens/ix-docs](https://github.com/siemens/ix-docs) — *N/A unless usage pages describe overlay-open trigger states*

## Summary by CodeRabbit

* **Bug Fixes**
  * Group context-menu triggers now remain visibly active while their menus are open.
  * Header avatar dropdown triggers now use updated subtle styling and clearly indicate their open state.
  * Active states and expanded accessibility indicators are cleared correctly when dropdowns close.
* **Tests**
  * Added regression and visual coverage for avatar and group dropdown open and close behavior.


packages/core/src/components/group/group-context-menu.tsx
```diff
@@ -65,10 +95,15 @@ export class GroupContextMenu {
     return (
       <Host>
         <ix-icon-button
-          class={{ hide: !this.showContextMenu }}
+          class={{
+            hide: !this.showContextMenu,
+            active: this.dropdownShow,
```
nuke-ellington, line 100: Looks like there is no styling asscociated with the `active` class (VRT did not change too). Is this intended?

nuke-ellington (changes requested):

packages/core/src/components/group/group-context-menu.tsx
alexkaduk: Yes — styling lives on ix-icon-button via :host(.active) / btn-active-look, not in group-context-menu. No local SCSS on purpose (same overlay-open pattern as date-dropdown / dropdown-button).

I refreshed the group context-menu Linux baselines. The delta is tiny (essentially the 24px kebab fill), but pixel-sampling matches subtle-tertiary Active (--theme-color-ghost--active) rather than Hover/Default once the pointer leaves the trigger.

nuke-ellington (approved):

Issue #2720, opened 2026-08-18
Collapsed Menu Category Hover Mouse behaviour


NiklasHae:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

If the `ix-menu` is in a collapsed state and someone hovers over an `<ix-menu-category>`, the items are shown in a popover during hover. Because there's a tiny gap between this popover and the menu item, moving your mouse slowly hides the popover, which makes it sometimes hard to click on the items.

I addded a screenshot with the gap:

<img width="368" height="175" alt="Image" src="https://github.com/user-attachments/assets/ff829e3a-a9cc-4a76-945d-860584d6f59d" />

I think it would be better if the gap is ignored or there's a bigger delay before hiding the menu.

## Diagonal mousemovement

Additionally, moving the mouse diagonally as indicated in the screenshot hides the menu as well:

<img width="383" height="168" alt="Image" src="https://github.com/user-attachments/assets/6055229a-4a82-414f-b9c6-3deec5640a06" />

This makes the popover hard to reach with the mouse sometimes.

### What type of frontend framework are you seeing the problem on?

Angular (Standalone)

### On which version of the frontend framework are you experiencing the issue?

angular@22

### Which version of iX do you use?

main

### Code to produce this issue.

```shell
It's reproduciable in the test-app in `/preview/menu-category`
```

flxlst09:
We already have related issues and will be working on it. Internal reference: EIX-211

NiklasHae:
Thx, good to know 👍 Let me know if you need help.

HelenaMohnatska:
I'm currently facing the same issue

Issue #2718, opened 2026-08-17
Add an option to disable ix-menu-item and ix-menu-category Tooltips


HelenaMohnatska:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### Suggestion / feature request

It would be useful to have an option to disable tooltips for menu items. For example, when the menu is expanded, displaying a tooltip is unnecessary. Having a dedicated property or configuration option to control tooltip visibility would provide better flexibility.

<img width="925" height="763" alt="Image" src="https://github.com/user-attachments/assets/cd23fce2-3b9d-4b57-b9fd-f16b32a9fb1f" />

flxlst09:
Internal reference: EIX-211

Issue #2644, opened 2026-07-08
Application menu - second level navigation doesn't always shown if hovered


paskaran:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

If app bar is on collapsed state try to move the mouse over the second level navigation. It appears sometime and most of the time its not visible. 
This issue is reproducible in our application aand also on your documentation of the Application Menu page: https://ix.siemens.io/docs/components/application-menu/code#2nd-navigation-level 

https://github.com/user-attachments/assets/3a766429-a4ad-4c9f-a802-585aab45c63e

### What type of frontend framework are you seeing the problem on?

Angular (NgModule)

### On which version of the frontend framework are you experiencing the issue?

21.2.13

### Which version of iX do you use?

v5.0.0

### Code to produce this issue.

```shell
See https://ix.siemens.io/docs/components/application-menu/code#2nd-navigation-level
```

chencc1988:
Duplicate with [this](https://github.com/siemens/ix/issues/2587)

flxlst09:
Internal reference EIX-211

Issue #2587, opened 2026-06-08
2nd level navigation items close unexpectedly


chencc1988:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

There is a gap between  the 1st level nav bar and 2nd level navbar
When the mouse moves from the 1st level to the 2nd level, the 2nd level items disappear.

<img width="345" height="177" alt="Image" src="https://github.com/user-attachments/assets/2a9de5b0-831d-4076-b0a7-5dca4c5396dd" />

<img width="511" height="357" alt="Image" src="https://github.com/user-attachments/assets/ef14ebe8-71c0-49b3-9f2c-7d99a0b51c53" />

### What type of frontend framework are you seeing the problem on?

Angular (Standalone)

### On which version of the frontend framework are you experiencing the issue?

 @angular/core@20.0.0

### Which version of iX do you use?

v5.0.0

### Code to produce this issue.

```shell
slowly move the mouse from 1st level to 2nd level
```

flxlst09:
Internal reference EIX-211

PR #2746, opened 2026-08-27
feat(menu): keep collapsed category popovers reachable and add disableTooltip EIX-211

alexkaduk:

When `ix-menu` is collapsed, hovering `ix-menu-category` opens nested items in an `ix-dropdown` (`placement="right-start"`, `offset={{ mainAxis: 3 }}`). `pointerleave` on the category host hid that popover immediately. The 3px visual gap between the rail and the panel is not part of the host or the dropdown, so a slow horizontal or diagonal move over the application chrome closed the popover before nested items could be reached.

`ix-menu-item` always rendered an `ix-tooltip`. There was no public way to turn it off. `ix-menu-category` had no matching control for the category parent.

GitHub Issue Number: #2718, #2720, #2587, #2644 (EIX-211)

- Collapsed **ix-menu-category** popovers stay open while the pointer crosses the 3px gap or moves diagonally toward nested items.
  - Transparent `::before` hit strip on `.category-dropdown` (4px = 3px Floating UI offset + 1px rounding; CSS pixels, both `right` and `left` placements).
  - 250ms hide delay, cancelled on re-enter of the host or dropdown.
  - `relatedTarget` “still inside” check covers host, host shadow, dropdown, and dropdown shadow (`host.contains()` misses shadow descendants).
  - Immediate hide is unchanged for `closeOtherCategories`, nested item click, and unmount. Keyboard / focus / Escape paths are unchanged. The visual 3px offset is unchanged.
- New **disableTooltip** prop (default `false`, `@since 6.0.0`) on **ix-menu-item** and **ix-menu-category**. When `true`, `ix-tooltip` is not rendered. Category forwards the prop to its inner **ix-menu-item**. Tooltips are not auto-hidden when the menu is expanded.
- Preview examples set `disableTooltip` / `disable-tooltip` on **Top level Category** only; **Normal Tab** still shows the default tooltip.

### Tests added

- CT: gap crawl (400ms pause in the strip so delay-alone cannot pass), diagonal move, nested-item selection, `disableTooltip` on category parent and on `ix-menu-item`.
- Existing CT coverage kept: sibling blur still closes after the delay, keyboard Space/arrows/Escape, collapsed dropdown open.

## 🏁 Checklist

- [ ] 🦮 Accessibility (a11y) features were implemented — *Existing CT a11y + Storybook menu a11y still pass; no new ARIA pattern*
- [ ] 📕 Add or update a Storybook story — *N/A (no new story; existing menu stories pass a11y)*
- [ ] 📄 Documentation was reviewed/updated [siemens/ix-docs](https://github.com/siemens/ix-docs) — *Follow-up: document **disableTooltip** on menu item / category in siemens/ix-docs after this ships in v6*
- [ ] 🧪 Unit tests were added/updated and pass (`pnpm test`) — *No `.spec.ts` for these components; CT added/updated*
- [ ] 📸 Visual regression tests were added/updated and pass ([Guide](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md#visual-regression-testing)) — *Existing `tests/menu/menu.e2e.ts` passed in Docker; no snapshot update*

## 👨‍💻 Help & support

- Generated Stencil wrappers (`packages/angular`, `packages/react`, `packages/vue`, `component-api.json`, `components.d.ts`) are part of this commit; do not hand-edit them.
- Inspector: the hit strip is `ix-menu-category` → `#shadow-root` → `ix-dropdown.category-dropdown` → `::before`. The popover must be open (`display: none` otherwise).
- Manual preview: `packages/react-test-app` → `/preview/menu-category`.

## Summary by CodeRabbit

- **New Features**
  - Added an option to disable tooltips for menu categories and menu items.
  - Added scrolling support for menu category popovers with many nested items.

- **Bug Fixes**
  - Improved collapsed menu category popovers so they remain open while moving across gaps or toward nested items.
  - Improved access to second-level menu items during pointer navigation.
  - Expanded the interactive area between categories and their popovers for smoother navigation.


flxlst09 (dismissed):
Behavior now matches the requirements, tested in Storybook. LGTM!

packages/core/src/components/menu-category/menu-category.scss
```diff
@@ -60,7 +60,28 @@
     display: none;
   }
 
+  // Hit-area over the 3px Floating UI offset (mainAxis: 3). Use px, not rem:
+  // the offset is CSS pixels. 4px = 3px gap + 1px for subpixel rounding.
   .category-dropdown {
+    overflow: visible;
```
danielleroux: Modify the overflow behaviour results into that all items are now visible:

Image

danielleroux (changes requested):

packages/core/src/components/menu-category/menu-category.scss
alexkaduk: `overflow: visible` on the host overrode ix-dropdown’s `overflow-y: auto`, so tall flyouts showed every item. 

That override is gone. The hit strip stays on the host (suppressOverflowBehavior); items scroll in .category-dropdown-body with the same 50vh - 3rem cap.

Might it work?

alexkaduk: The compact basic md snapshot is valid at 768px — 50vh - 3rem never kicks in there. Storybook sets html, body { height: 85vh }, so the same cap is smaller than the flyout. The column flex then shrinks the header and the body shows a scrollbar. Also, the title used ix-dropdown-item with the empty checkmark column, which extra-indented the label.

Fix: height: fit-content, header flex: 0 0 auto, suppressChecked, body flex: 0 1 auto. Tall lists still scroll. 
Added a short-viewport CT and visual snapshot.

lzeiml (approved):

dmytro-halimov (approved):

PR #2786, opened 2026-09-10
fix: component hardening

danielleroux:

Tracked: EIX-166

## 🏁 Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Space characters can now be entered in category filter inputs without triggering dropdown actions.
  - Reserved tree item IDs work correctly without affecting other application data.
  - Modal configuration now applies only supported options.

- **Security**
  - Unsafe navigation protocols are blocked across buttons, links, menus, breadcrumbs, and application switching.
  - External applications open without access to the originating window.

- **Documentation**
  - Added category filter examples and clarified supported link URL formats.

### DIFF
--- a/packages/core/src/components/application-header/application-switch-modal/application-switch-modal.tsx
+++ b/packages/core/src/components/application-header/application-switch-modal/application-switch-modal.tsx
@@ -14,6 +14,7 @@ import {
 } from '../../utils/application-layout/context';
 import { dismissModal } from '../../utils/modal';
 import { iconApps, iconOpenExternal } from '@siemens/ix-icons/icons';
+import { getSafeNavigationUrl } from '../../utils/condition-checks';
 
 function ApplicationItem(props: {
   host: HTMLIxApplicationSwitchModalElement;
@@ -50,8 +51,16 @@ function ApplicationItem(props: {
         Selected: props.selected,
       }}
       onClick={() => {
+        const url = getSafeNavigationUrl(
+          props.url,
+          'ix-application-switch-modal'
+        );
+        if (url === undefined) {
+          return;
+        }
+
         dismissModal(props.host);
-        window.open(props.url, props.target);
+        window.open(url, props.target, 'noopener');
       }}
     >
       <img class="AppIcon" src={props.iconSrc} alt=""></img>
--- a/packages/core/src/components/application-header/application-switch-modal/application-switch-modal.tsx
+++ b/packages/core/src/components/application-header/application-switch-modal/application-switch-modal.tsx
@@ -60,7 +60,11 @@ function ApplicationItem(props: {
         }
 
         dismissModal(props.host);
-        window.open(url, props.target, 'noopener');
+        if (isExternal(props.target)) {
+          window.open(url, props.target, 'noopener');
+        } else {
+          window.open(url, props.target);
+        }
       }}
     >
       <img class="AppIcon" src={props.iconSrc} alt=""></img>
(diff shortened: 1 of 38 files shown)


packages/core/src/components/utils/uuid.ts
```diff
@@ -7,14 +7,7 @@
...
 export function generateUUID(): string {
-  const randomString =
-    Math.random().toString(36).substring(2, 15) +
-    Math.random().toString(36).substring(2, 15);
-
-  const timestampString = new Date().getTime().toString(36);
-  const uuid = randomString + timestampString;
-
-  return uuid;
+  return crypto.randomUUID();
```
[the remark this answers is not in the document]
alexkaduk: pls check this. might worth to adjust

packages/core/src/components/application-header/application-switch-modal/application-switch-modal.tsx
```diff
@@ -50,8 +51,16 @@ function ApplicationItem(props: {
...
+          props.url,
+          'ix-application-switch-modal'
+        );
+        if (url === undefined) {
+          return;
+        }
+
         dismissModal(props.host);
-        window.open(props.url, props.target);
+        window.open(url, props.target, 'noopener');
```
alexkaduk: setting `noopener` discards the target name and always opens a fresh _blank context. AppSwitchConfigurationTarget allows _self, _parent, _top, and arbitrary window names — the new test in this PR uses target: 'reports'. 

So an app configured with _self now spawns a new window instead of navigating the current one. Could we apply noopener only when the target is _blank or a custom name?

packages/core/src/components/button/base-button.tsx
alexkaduk: This branches on the raw props.href, so a rejected URL renders <a role="button" tabindex="0"> with no href: focusable, announced as a button, inert on Enter. Branching on the sanitized href from line 102 would fall back to a real <button> instead.

packages/core/src/components/menu-item/menu-item.tsx
alexkaduk: Same as base-button: branching on the raw this.href leaves a focusable anchor with no href when the URL is rejected. Using the sanitized href from line 318 would fall back to the <button> branch.

packages/core/src/components/link-button/link-button.tsx
alexkaduk: The raw URL is still surfaced as the tooltip, so a blocked javascript:… string stays visible to users even though href was dropped. Combined with the hardcoded tabindex="0" on the next line, this leaves a focusable link that does nothing.

packages/core/src/components/tree/tree.tsx
```diff
@@ -202,19 +202,22 @@ export class Tree {
 
   private getContext(id: string): TreeItemContext {
     if (!this.context) {
-      return {
-        isExpanded: false,
-        isSelected: false,
-        isDisabled: this.model[id]?.disabled,
-      };
+      this.context = {};
```
alexkaduk: Two small things: Object.create(null) would make the map immune by construction rather than relying on the hasOwnProperty guard below. And context is @Prop({ mutable: true }), so when a consumer passes null this now assigns a prop during render() — the previous code returned a throwaway object instead.

packages/core/src/components/tree/tree.tsx
```diff
@@ -32,6 +32,8 @@ import {
 } from './tree-model';
 import { defaultRefreshTreeOptions, RefreshTreeOptions } from './tree.types';
 
+const createTreeContext = (): TreeContext => Object.create(null);
```
nuke-ellington: Default and emitted contexts now have a null prototype. Existing consumer code such as  `context.hasOwnProperty(id)`  or  `context.toString()`  will throw. The reserved-ID protection can retain ordinary objects by using own-property checks and  `Object.defineProperty()`  without converting the public context. Otherwise, this should be treated and documented as a breaking change rather than a patch.

nuke-ellington (changes requested):

Issue #2780, opened 2026-09-09
Min- and Max-Date in Date Input Issue


6axter82:
### Prerequisites

- [x] I have read the [Contributing Guidelines](https://github.com/siemens/ix/blob/main/CONTRIBUTING.md).
- [x] I have not leaked any internal/restricted information like screenshots, videos, code snippets, links etc.

### What happened?

Our team is using IX Version 4. I have checked, the issue is also present in the IX Version 5 of the Date Input Component.

When setting the min and max dates, and clicking on a date input dropdown, the calendar view is shown. Clicking the arrows left and right brings to the possible selected month and dates, reaching the min and max explicitly specified dates.
But clicking on the month-name-button between the arrows and trying to select the min date month is not possible, only the months coming after the minDate month are selectable. Also, the maxDate is not limited as it should be, the next month after the maxDate month is also available for selection in this view.

Please correct the issue. If possible also in version 4 as well.

I have run our codes example and explicitly placed the minDate=2026/03/31 and maxDate=2027/01/31.
Can confirm the bug. It is always one month after the limited month.

Thank you in advance,
Aziz Dossanov

### What type of frontend framework are you seeing the problem on?

Angular (Standalone)

### On which version of the frontend framework are you experiencing the issue?

@angular/core@21.2.19

### Which version of iX do you use?

v4.4.0, but also checked the v5

### Code to produce this issue.

```shell
https://ix.siemens.io/version-4.4.0/docs/components/input-date/code
```

flxlst09:
Thank you @6axter82 for providing the information. We added this issue to our backlog, internal reference EIX-314

PR #2822, opened 2026-09-17
Fix(core date input) month list disable offset eix 314

benjgil:

There are numerous bugs in the code related to mismatched indexes being compared across the logic. E.g. the month disabling code that compares 0-base index month names to 1-base month list.

GitHub Issue Number: #2780 

This PR fixes any existing Date index mismatches and removes the scope for any further mismatches via normalising all DateTime logic to use consistent Luxon objects and then resolve strings and values directly from these objects when required, rather than using the strings and values for comparison.

Tests have been added to ensure behaviour is kept consistent, and to cover the revealed prior bugs.

## 🏁 Checklist

## Summary by CodeRabbit

* **New Features**
  * Improved date-picker month and year navigation using more reliable calendar handling.
  * Added localized month and weekday labels with support for configurable week starts.
  * Improved enforcement of minimum and maximum date ranges for day, month, and year selection.
  * Added support for Sunday-first calendar layouts.

* **Bug Fixes**
  * Prevented invalid calendar selections and corrected date positioning across locales and week-start configurations.

* **Tests**
  * Added comprehensive coverage for calendar calculations, date ranges, localization, leap years, and week layouts.

### DIFF
--- a/packages/core/eslint.config.cjs
+++ b/packages/core/eslint.config.cjs
@@ -17,6 +17,47 @@ const compat = new FlatCompat({
   allConfig: js.configs.all,
 });
 
+// `no-restricted-syntax` is not additive across flat-config objects - a later
+// object replaces the whole list. Keep the shared entries here so the scoped
+// override below can re-declare them alongside its own.
+const noAssertionSyntaxRules = [
+  {
+    selector:
+      "TSNonNullExpression:not([expression.type='MemberExpression'][expression.property.name='shadowRoot']):not([expression.type='CallExpression'][expression.callee.type='MemberExpression'][expression.callee.property.name='querySelector'])",
+    message:
+      'Avoid non-null assertion (!). Use safer narrowing, except for shadowRoot access and querySelector(...) results.',
+  },
+  {
+    selector:
+      "PropertyDefinition[definite=true]:not(:has(Decorator[expression.callee.name='Event'])):not(:has(Decorator[expression.callee.name='Prop'])):not(:has(Decorator[expression.callee.name='Element'])):not(:has(Decorator[expression.callee.name='Element'])):not(:has(Decorator[expression.callee.name='AttachInternals']))",
+    message:
+      'Avoid definite assignment assertion (!). Initialize the field or use an allowed decorator field.',
+  },
+];
+
+// Luxon counts months 1-12 and weekdays 1-7 (Monday = 1), but the name arrays
+// from `Info.months()` / `Info.weekdays()` and the pickers' own state are
+// 0-based. Mixing the two silently shifts the calendar by one unit, so the
+// date components must go through `utils/calendar-units.ts`, which owns every
+// conversion.
+const calendarUnitRules = [
+  {
+    selector: "MemberExpression[property.name='month']",
+    message:
+      "Luxon's .month is 1-12 but the UI month arrays are 0-based. Use monthIndexOf() / fromMonthIndex() from utils/calendar-units.ts.",
+  },
+  {
+    selector: "MemberExpression[property.name='weekday']",
+    message:
+      "Luxon's .weekday is 1-7 (Monday = 1) but the UI weekday arrays are 0-based. Use weekdayIndexOf() / weekdayColumnOf() from utils/calendar-units.ts.",
+  },
+  {
+    selector: "NewExpression[callee.name='Date'][arguments.length>=2]",
+    message:
+      'new Date(year, month, day) takes a 0-based month and resolves in local time. Use fromMonthIndex() from utils/calendar-units.ts.',
+  },
+];
+
 module.exports = [
   {
     ignores: [
@@ -52,24 +93,27 @@ module.exports = [
       'react/jsx-uses-react': 0,
       'react/react-in-jsx-scope': 0,
       '@typescript-eslint/no-confusing-non-null-assertion': 'error',
-      'no-restricted-syntax': [
-        'error',
-        {
-          selector:
-            "TSNonNullExpression:not([expression.type='MemberExpression'][expression.property.name='shadowRoot']):not([expression.type='CallExpression'][expression.callee.type='MemberExpression'][expression.callee.property.name='querySelector'])",
-          message:
-            'Avoid non-null assertion (!). Use safer narrowing, except for shadowRoot access and querySelector(...) results.',
-        },
-        {
-          selector:
-            "PropertyDefinition[definite=true]:not(:has(Decorator[expression.callee.name='Event'])):not(:has(Decorator[expression.callee.name='Prop'])):not(:has(Decorator[expression.callee.name='Element'])):not(:has(Decorator[expression.callee.name='Element'])):not(:has(Decorator[expression.callee.name='AttachInternals']))",
-          message:
-            'Avoid definite assignment assertion (!). Initialize the field or use an allowed decorator field.',
-        },
-      ],
+      'no-restricted-syntax': ['error', ...noAssertionSyntaxRules],
       '@typescript-eslint/no-explicit-any': 'warn',
       '@typescript-eslint/no-unused-vars': 'warn',
       'no-unused-vars': 'off',
     },
   }),
+  {
+    files: [
+      'src/components/date-picker/**/*.{ts,tsx}',
+      'src/components/date-input/**/*.{ts,tsx}',
+      'src/components/date-dropdown/**/*.{ts,tsx}',
+      'src/components/datetime-picker/**/*.{ts,tsx}',
+      'src/components/datetime-input/**/*.{ts,tsx}',
+    ],
+    ignores: ['**/test/**', '**/tests/**'],
+    rules: {
+      'no-restricted-syntax': [
+        'error',
+        ...noAssertionSyntaxRules,
+        ...calendarUnitRules,
+      ],
+    },
+  },
 ];
--- a/packages/core/src/components/date-picker/date-picker.tsx
+++ b/packages/core/src/components/date-picker/date-picker.tsx
@@ -26,6 +26,18 @@ import {
 } from '@stencil/core';
 import { DateTime, Info } from 'luxon';
 import type { DateTimeCardCorners } from '../date-time-card/date-time-card.types';
+import {
+  type MonthIndex,
+  addMonths,
+  daysInMonth,
+  fromMonthIndex,
+  isDayWithinRange,
+  isMonthWithinRange,
+  isYearWithinRange,
+  localFromMonthIndex,
+  monthIndexOf,
+  weekdayColumnOf,
+} from '../utils/calendar-units';
 import { queryElements } from '../utils/focus/focus-utilities';
 import { DefaultMixins } from '../utils/internal/component';
 import { makeRef } from '../utils/make-ref';
@@ -340,9 +352,7 @@ export class DatePicker
   }
 
   private getDaysInCurrentMonth(): number {
-    return (
-      DateTime.utc(this.selectedYear, this.selectedMonth + 1).daysInMonth || 0
-    );
+    return daysInMonth(this.selectedYear, this.selectedMonth);
   }
 
   private getFirstDayOfWeek(day: number): number {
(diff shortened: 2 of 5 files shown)


alexkaduk:
This is a user-facing month min/max bug fix, so it needs a changeset for @siemens/ix (patch) or it will not ship in release notes.

packages/core/src/components/date-picker/date-picker.tsx
```diff
@@ -428,12 +454,10 @@ export class DatePicker
    */
   @Method()
   async updateSelectedYearMonth(date: DateTime) {
```
alexkaduk, line 456: `renderMonths()` builds the list from `tempMonthDate.year`, but `updateSelectedYearMonth` only assigns `selectedMonthDate`.

`from`/`to` watchers and `ix-date-dropdown` (re-select same range) call this method, so the calendar can show year A while the month dropdown still lists/disables year B.

Please also set `this.tempMonthDate = this.selectedMonthDate` here, same as `selectMonth` / `changeCalendarView`.

packages/core/src/components/date-picker/test/date-picker.ct.ts
```diff
@@ -645,6 +645,83 @@ regressionTest.describe('keyboard navigation', () => {
...
+        /disabled-item/
+      );
+    }
+  };
+
+  regressionTest(
+    'enables the month a single-month range sits in',
+    async ({ mount, page }) => {
+      await mount(
+        `<ix-date-picker from="2026/07/06" min-date="2026/07/05" max-date="2026/07/15" single-selection></ix-date-picker>`
```
alexkaduk, line 681: These asserts use English month names (`July`, `June`, …) but the mounts do not set `locale="en"`. The translation test in this file already pins `locale="de"`, so this will fail on a non-English runner.

Please add `locale="en"` to these mounts. A case matching #2780 (`min-date="2026/03/31"` / `max-date="2027/01/31"`) would also help.

alexkaduk:
The **PR title** (`Fix(core date input) month list disable offset eix 314`) does not match `<type>[optional <scope>]: <description>` from `.github/instructions/code-review.instructions.md` (same rule in `AGENTS.md`).

Please rename the pull request to something like `fix(core/date-picker): correct month dropdown min/max disable offset EIX-314`.

Also please rebase onto latest `main`. #2767 extracted `DatePickerYearMonth` after this branch’s last merge; this PR types `updateSelectedYearMonth` as `DateTime` again, which is the right call now that month is no longer a bare number — just avoid leaving the unused type behind.

packages/core/src/components/date-picker/date-picker.tsx
```diff
@@ -428,12 +454,10 @@ export class DatePicker
    */
   @Method()
```
alexkaduk, line 455: The DateTime model is good! however the displayed month is still two clocks (`selectedMonthDate` vs `tempMonthDate`) written four different ways.

`renderMonths()` uses `tempMonthDate.year`, while the header and grid use `selectedMonthDate`. `updateSelectedYearMonth` (from/to watchers and `ix-date-dropdown`) only sets `selectedMonthDate`, so the calendar can show year A while the month list still disables year B. The JSDoc says `tempMonthDate` diverges only while picking a year before a month, but year click already commits both.

Pls consider to route all four paths through one setter, e.g. `setDisplayedMonth(month)`, that assigns both to `month.startOf('month')`. Keep the current UX (year click updates the calendar immediately). Then the `tempMonthDate` comment should match that, or go away.

alexkaduk: It is related to the next comment. Pls consider both as one suggestion/finding

packages/core/src/components/utils/calendar-units.ts
```diff
@@ -0,0 +1,179 @@
+/*
+ * SPDX-FileCopyrightText: 2026 Siemens AG
+ *
+ * SPDX-License-Identifier: MIT
+ */
+
+import { DateTime, Info } from 'luxon';
+
+/**
```
alexkaduk, line 9: A lot of the new commentary restates the 0- vs 1-based rule (file essay, almost every helper, ESLint block, and `tempMonthDate` / `weekStart` in `date-picker.tsx`). That is heavier than nearby core utils (`date-time-locale.ts` is a short JSDoc only when the contract is not obvious).

Please keep public `@Prop`/`@Method` docs, one short module note, and the ESLint `message` strings. Trim the rest — including the `tempMonthDate` JSDoc if we keep a single `setDisplayedMonth` (it currently describes a year-then-month flow that year click does not do). Also the `locale` prop still says locale sets weekday order; `weekStartIndex` does.

alexkaduk:
`ix-date-input` exposes `weekStartIndex` but does not pass it to the nested `ix-date-picker` (`date-input.tsx`, the `<ix-date-picker>` around line 608). `ix-date-dropdown` and `ix-datetime-picker` already forward it, so on date-input the public prop is a no-op (week always starts Monday).

Please consider to add `weekStartIndex={this.weekStartIndex}` on that nested picker.

packages/core/eslint.config.cjs
```diff
@@ -17,6 +17,59 @@ const compat = new FlatCompat({
   allConfig: js.configs.all,
 });
 
+// `no-restricted-syntax` is not additive across flat-config objects - a later
```
alexkaduk, line 20: Reading `eslint.config.cjs`: this PR extracts the existing `!` / definite-assignment `no-restricted-syntax` rules so they can be re-spread, then adds `calendarUnitRules` on the date-* components (ban `.month`, `.weekday`, `Info.months()` / `Info.weekdays()`, `new Date(y, m, …)`; tests and `calendar-units.ts` excluded).

That is a guardrail so we do not mix Luxon 1-based ordinals with 0-based name arrays again. What I see it is not required for the month min/max fix — `calendar-units` + the picker DateTime state already do that. Flat config replaces `no-restricted-syntax` rather than merging, which is why the `!` rules had to be extracted.

CI should not fail if this file is reverted on this PR: core lint is still the `main` `!` rules, and the new picker code does not trip them.

Please move the ESLint change to a follow-up/new PR, so this PR stays the #2780 / EIX-314 behaviour fix.

packages/core/src/components/date-picker/test/date-picker.ct.ts
benjgil: Added
The specific date range from #2780 isn't so relevant as a test case - the issue there was the month index being incorrect, the specific date range is a red herring. 
But it's worth adding a range that spans a year, so that's been added

packages/core/src/components/date-picker/date-picker.tsx
benjgil: Its leftovers from many years of date-picker changes, just removed the tempMonthDate now

packages/core/src/components/date-picker/date-picker.tsx
benjgil: removed tempMonthDate

packages/core/src/components/utils/calendar-units.ts
benjgil: trimmed

packages/core/eslint.config.cjs
benjgil: Disagree - it's required so that this sort of fix isn't needed again, so it's very relevant to the entire refactor this PR implements
(and eslint rules are by definition added as future guardrails - they should never fail on CI of the branch in which they were added...)

packages/core/src/components/date-picker/date-picker.tsx
```diff
@@ -428,13 +453,12 @@ export class DatePicker
...
-    this.selectedMonth = date.month - 1;
+  async updateSelectedYearMonth(date: DateTime) {
+    this.selectedMonthDate = date.startOf('month');
+    this.tempMonthDate = this.selectedMonthDate;
   }
 
-  @Watch('selectedYear')
-  @Watch('selectedMonth')
+  @Watch('selectedMonthDate')
   @Watch('weekStartIndex')
```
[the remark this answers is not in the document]
benjgil: done

packages/core/src/components/date-picker/test/date-picker.ct.ts
```diff
@@ -645,6 +645,167 @@ regressionTest.describe('keyboard navigation', () => {
...
+
+      await expectMonths(
+        monthSelection,
+        ['July', 'August'],
+        ['June', 'September']
+      );
+    }
+  );
+
+  // #2780: a range spanning a year boundary must only constrain the months of
```
alexkaduk: Would suggest to remove this id. No sure if comment is needed. The test description should cover details.

packages/core/src/components/date-dropdown/test/date-dropdown.ct.ts
```diff
@@ -248,6 +248,69 @@ regressionTest('select different year', async ({ mount, page }) => {
   await expect(monthContainer).toHaveText(/March/);
 });
 
+regressionTest(
+  're-selecting the same range moves the month dropdown back to the range year',
+  async ({ mount, page }) => {
+    await mount(`<ix-date-dropdown></ix-date-dropdown>`);
```
alexkaduk: This re-select test asserts the English month name `February`, but the mount does not set `locale="en"`. Month labels follow the runner locale when the prop is unset (same issue as the date-picker month tests).

Please add `locale="en"` on this `ix-date-dropdown`.

packages/core/eslint.config.cjs
alexkaduk: @danielleroux could you pls check eslint config updates part and suggest?

packages/core/src/components/date-picker/test/date-picker.ct.ts
benjgil: agreed

packages/core/src/components/date-dropdown/test/date-dropdown.ct.ts
benjgil: this makes sense for total completeness, but where do you envisage runners not having locale="en"?

packages/eslint-config-ix/rules/no-luxon-calendar-ordinals.js
```diff
@@ -0,0 +1,167 @@
...
+      const variable = sourceCode
+        .getScope(node)
+        .references.find((reference) => reference.identifier === node.callee);
+
+      return (
+        !variable || !variable.resolved || variable.resolved.defs.length === 0
+      );
+    };
+
+    return {
```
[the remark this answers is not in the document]
benjgil: done

packages/core/src/components/date-dropdown/test/date-dropdown.ct.ts
benjgil: (implemented)

packages/core/eslint.config.cjs
danielleroux: I see the value in preventing another ordinal mismatch, but I don’t think this ESLint rule is the best primary safeguard. It bans particular syntax in five component directories, it cannot verify that the calendar is correct, and equivalent code such as `dateTime.get('month')` can bypass it. It also rejects some safe uses of `.month`. So it enforces a coding convention, not the behavior we need.

I’d rather make `calendar-units.ts` the single place that handles month and weekday conversions, and move the remaining grid calculation there too. Its API should accept dates and return month options and calendar rows, so components do not have to combine 1-based Luxon values, 0-based indexes, and ISO week numbers themselves. Comments can explain those boundaries. Tests can enforce the outcomes.

benjgil: I dont see any reason to not have a lint safeguard for the major use cases - primary or not, it helps reduce the potential for index mis-match to creep in. 

Good point about moving calendar calculation into calendar-units.ts (all other logic was already there) so that we end this change with everything under one roof, but this doesn't prevent people (or agents) from adding new logic that bypasses that file. 

But I still see value in linting safeguards, which dont reject safe uses of .month - they're limited to only Luxon objects.

For agents, i'll add a note to the agents.md file that all datetime object logic should be added in calendar-units.ts, which should help with agent-based changes (let's face it, that's going to be 99% of updates) but still not fool-proof

benjgil: Added directions in agents.md and moved calendar calculation to calender.util.ts (new name for calendar-units.ts) - @danielleroux please review again

